PR #1, opened 2011-04-30
Made a bunch of clean up changes

tneil:
Changes consist of 
- File renames
- TOC changes
- Addition of class filters

General clean-up work


tneil:
General changes look good.. these clean up some of the rendering artifacts seen in the initial source base

PR #2, opened 2011-05-02
namespace, class, and method cleanup for generated docs

marek:
- renamed menu/menuitem files to fall under blackberry_ui_
- fixed overload for blackberry.push.Data.prototype.getHeaderField() so it does not get clobbered as a duplicate
- removed duplicate blackberry.invoke.BrowserArguments


api/blackberry_push_Data.js
```diff
@@ -120,16 +113,16 @@ blackberry.push.Data.DECLINE_USERRFS = 5;
 * @returns {String}
 * @BB50+
 */
-blackberry.push.Data.prototype.getHeaderField = function(field) { };
-
 /**
 * Returns the value of a header field. 
+* @name blackberry.push.Data.prototype.getHeaderField^2
```
jeffheifetz, line 121: This should not be necessary. Removing the duplicated set of comments and adding an @function tag should fix the issue.

api/blackberry_push_Data.js
```diff
@@ -120,16 +113,16 @@ blackberry.push.Data.DECLINE_USERRFS = 5;
...
-blackberry.push.Data.prototype.getHeaderField = function(field) { };
-
 /**
 * Returns the value of a header field. 
+* @name blackberry.push.Data.prototype.getHeaderField^2
 * @param {Number} field This parameter is the integer index of the header to retrieve.
 * @returns {String}
 * @BB50+
 */
 blackberry.push.Data.prototype.getHeaderField = function(field) { };
```
jeffheifetz, line 126: While you're fixing the file, if you remove the prototype from the function names, they render properly with # characters

api/blackberry_push_Data.js
marek, line 121: They are not duplicate, but overloaded.  One takes a string, the other takes a number.

jeffheifetz, line 121: Nice catch, we'll have to find a uniform way to document this

api/blackberry_push_Data.js
jeffheifetz, line 126: I never realized this was to differentiate instance and static children. I guess we should just leave it

jeffheifetz:
Still have the empty namespace for blackberry.ui

jeffheifetz:
I think its great to go

PR #3, opened 2011-05-02
Dynamic js files fix

jeffheifetz:
Now we can actually run jsdoc without it crashing.


cpinzon:
Looks good, especially the first commit (72bbc09) will prevent the return of an error message claiming that BBTemplate is not a valid template

PR #4, opened 2011-05-02
Adding DedicatedWorker API to js doc.

tlai:


tlai:
We don't support SharedWorkerGlobalScope. I will change the toc and name.

PR #6, opened 2011-05-02
Learn tag

cpinzon:
This feature branch adds support for a learn tag. 

The learn tag is supposed to notify the developer of learning resources available for the specific API that they are viewing. It would be added in the following manner: 

/**
- @namespace The Utils object provides useful utility functions and properties.  
- @toc {Utilities} Utils
- @featureID blackberry.utils
- @learns {W3C Schools} http://www.w3schools.com/ This link redirects to W3C Schools, where they offer free tutorials in all web development technologies.
  */

I will send a picture of what the addition of the tag to an API would look like, as I seem unable to figure how to add a picture to this discussion.

### DIFF
--- a/build/BBTemplate/class.tmpl
+++ b/build/BBTemplate/class.tmpl
@@ -86,6 +86,24 @@
 				vertical-align: top;
 				background-color: #FCD8B9;
 			}
+			
+			table.learnTable {
+				border-width: 1px;
+				border-style: solid;
+				border-color: #ccc;
+				padding: 5px;
+				vertical-align: top;
+				margin: 5px;
+				margin-bottom:15px;
+				width : 90%;
+				background-color: #99FFCC;
+			}
+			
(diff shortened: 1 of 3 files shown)


tneil:
Hi Carolina... does the feature allow for multiple @learns tags?

build/BBTemplate/class.tmpl
```diff
@@ -93,6 +111,14 @@
 		<script src="server.js"></script>
 	</head>
 	
+
```
jeffheifetz, line 114: I think the description should go below the beta tag

cpinzon:
Tim, I've added a few changes to ensure multiple tags are supported.

jeffheifetz:
Does anyone else think the learn tags should go at the bottom?

cdelcol:
let's just keep them where they are for now. They can always be moved later.

PR #8, opened 2011-05-03
HTML5 Media 

marek:
Initial work for HTMLMediaElement


tneil:
Hi Marek.. is the intention to document the root level media element instead of the audio/video or to document it in addition to the audio/video elements?

marek:
It will be in addition to audio/video. They are both derived off it

PR #9, opened 2011-05-03
Completed namespaces for HTML5 database

jackluo-zz:
The first completed version of HTML5 database


jeffheifetz:
I think you also need to document window.openDatabase to show the method to get a Database object.

PR #12, opened 2011-05-04
DeviceMotion API changes

cpinzon:
Changes to HTML5 DeviceMotion API
- removal of use of @class in favour of namespace
- removal of use of @property tag in favour of the desired use of tags. 

Also added DeviceOrientation API.

I will forward the generated HTML files in a .zip file.


jeffheifetz:
We need to change the images to BB ones. I remember Franklyn had some PB images but I don't know where those are.

cpinzon:
Sure, should I try to replicate the drawings on the BB images similarly to the arrows and vectors on the android ones?

PR #13, opened 2011-05-04
Fixed the 5.0 audio api

tneil:
Updated the TOC so that it is in the same media section as the HTML5 media elements.  Fixed the jsdocs format.  Added missing documentation and placed examples with their proper functions


tneil:
Merging in changes so that updates are available for folks when they get in the office in the morning

PR #14, opened 2011-05-04
Various fixes

tneil:
updated jsdocs formatting structure, OS tagging, examples and links to HTML5 toolkit

### DIFF
--- a/api/blackberry_find_FilterExpression.js
+++ b/api/blackberry_find_FilterExpression.js
@@ -124,38 +124,42 @@
 *   }
 * &lt;&sol;script&gt;
 */
-blackberry.find.FilterExpression = function(leftField,operator,rightField,negate) { };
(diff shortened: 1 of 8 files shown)


jeffheifetz:
There is no need for support tags at the namespace level. To prevent errors the templates calculate the support based on an object's children.

tneil:
Cool.. One less thing to update :)

tneil:
We have been using @default for other constants, so I just followed that pattern :)

Do we want to use @returns instead?  I know that the @default is displaying properly in the HTML, I haven't tried @returns

ghost:
Tim, is this pull request ready for a merge?

tneil:
sure.. it is ready for merge.. I'm working on the Phone/Call logs now... but I will send it as a follow up pull request

ghost:
OK. I'm merging it right now to the master.

PR #16, opened 2011-05-04
Updated all of the Phone jsdocs

tneil:
The phone jsdocs should be all good to go!

### DIFF
--- a/api/blackberry_phone_Find_FilterExpression.js
+++ b/api/blackberry_phone_Find_FilterExpression.js
@@ -16,18 +16,10 @@
 
 
 /**
-* @toc {Phone} Find 
-* @BB50+
-* @namespace
-*/
-blackberry.phone.Find = { };
-
-
-/**
 * <div><p>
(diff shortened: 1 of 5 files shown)


jeffheifetz:
The reason why Marek had all those .prototypes is because it differentiates between static and instance methods. Do we not want that designation?

tneil:
no problem... didn't know that was the reason.. i can change them back.  i added a bunch of updated descriptions and toc changes as well that i will keep

tneil:
Ok.. I changed the structure back to the way it was before to denote static and member functions/properties.  Have the templates been updated so that you don't get the hash #?  right now anything that uses 'prototype' has a # instead of a '.' before the function/property name.  

I would assume that it is supposed to be a '.' and we place the word 'static' in front of the function to denote that it is a static function or property

jeffheifetz:
I will make a change in the templates to denote static with the keyword vs the hash

PR #35, opened 2011-05-06
@constructedBy attribute

marek:
Added @constructedBy to document how objects are created that do not use the regular constructor.
For example:

HTMLAudioElement is created by new Audio()
HTMLVideoElement is created by document.createElement("video");

if a @constructedBy is specified, and no @constructor is present, the default constructor is not generated.
if both @constructedBy and @constructor are present. Both are generated

/**
- this is a MyClass description
- @class
- @constructor
- @param {Number} [foo] foo is a constructor param
- @constructedBy {document.createElement("MyClass")} {This is a description}
- // this is an example of how to create a new instance of MyClass
- var instance = document.createElement("MyClass");
  */

MyClass = function(foo) { }

### DIFF
--- a/build/BBTemplate/class.tmpl
+++ b/build/BBTemplate/class.tmpl
@@ -454,15 +454,21 @@
     {+ResetIndex()+}
     <if test="data.constructedBy">
         <for each="constructedBy" in="data.constructedBy">
-            <h2><a name="constructedBy{+GenerateIndex()+}"> </a>{+constructedBy.type+}</h2>
-            <hr style="margin-bottom:15px;"/>
-			<div class="title">Description</div>
-			&nbsp;{+resolveLinks(constructedBy.desc)+}
-            <table class="scriptTable">				
-                <tr><td class="scriptTd">
-                <pre>{+constructedBy.example+}</pre>
-                </td></tr>
-            </table>
+            <if test="constructedBy.type">
+                <h2><a name="constructedBy{+GenerateIndex()+}"> </a>{+constructedBy.type+}</h2>
+                <hr style="margin-bottom:15px;"/>
+                <if test="constructedBy.desc">
+                    <div class="title">Description</div>
+                    &nbsp;{+resolveLinks(constructedBy.desc)+}
+                </if>
+                <if test="constructedBy.example">
+                    <table class="scriptTable">
+                        <tr><td class="scriptTd">
+                        <pre>{+constructedBy.example+}</pre>
+                        </td></tr>
+                    </table>
+                 </if>
+             </if>
         </for>
     </if>
 <!-- ============================== Function details ========================= -->		
--- a/build/BBTemplate/class.tmpl
+++ b/build/BBTemplate/class.tmpl
@@ -245,11 +245,21 @@
 
 <!-- ============================== Constructor Summary ========================= -->
 <!-- NEW: BETA -->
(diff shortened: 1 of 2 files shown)


build/bbPlugin.js
```diff
@@ -33,7 +66,9 @@ JSDOC.PluginManager.registerPlugin(
                 var betaTag = symbol.comment.getTag("beta");
                 var paramCallbacks = symbol.comment.tags.filter(function($){return $.isCallback && $.title == "param"});
                 var fieldCBs = symbol.comment.tags.filter(function($){return $.isCallback && $.title == "field"});
-		var learnTag = symbol.comment.getTag("learns");
+                var learnTag = symbol.comment.getTag("learns");
+                var constructorTag = symbol.comment.getTag("constructor");
+                var constructedBy = symbol.comment.tags.filter(function($){return $.title=="constructedBy" && !!$.type});
```
jeffheifetz: Is that a double negative?

build/bbPlugin.js
```diff
@@ -126,9 +169,18 @@ JSDOC.PluginManager.registerPlugin(
...
-		}
+                if (docTag.title == "learns"){
+                    docTag.desc = docTag.nibbleName(docTag.desc);
+                }
+                if(docTag.title == "constructedBy") {
+                    if(!docTag.type) {
+                        LOG.warn(docTag.title + " missing type");
+                    } else {
+                        var parts = GetType(docTag.desc);
+                        docTag.desc = parts.type;
```
jeffheifetz, line 190: The tag should already be parsed. You should not need to do so again

build/BBTemplate/class.tmpl
```diff
@@ -436,6 +446,20 @@
 			</if>
 		</div>
 	</if>
+    {+ResetIndex()+}
```
jeffheifetz, line 454: I cannot make this call from BBPlugin. If you want to be able to you'll need to declare these in the publish.js

build/bbPlugin.js
marek: http://navirudra.com/blog/javascript/javascript-double-negative-trick.html

build/bbPlugin.js
marek, line 190: This parses two sets. jsDocs only parses one by default.

build/bbPlugin.js
jeffheifetz: Nice trick, Thanks! I never knew about this.
If it covers additional cases, should we throw this everywhere?

build/bbPlugin.js
jeffheifetz, line 190: Sorry didn't notice

build/BBTemplate/class.tmpl
jeffheifetz, line 454: Got it working with our help, thanks

jeffheifetz:
Seems like this should be good enough for now, we can always expand it in the future if we need different platform support per constructor

jeffheifetz:
Is there a way to set different sets of parameters for each constructor?

marek:
Nope there is no way of defining the parameters. 
Felt that it would be odd to describe the parameter types for a call like document.createElement("audio"); when it's a fixed string you use.

jeffheifetz:
How will Jack use it for something like window.openDatabase that has like 4 parameters?

----- Original Message -----

marek:
It's limited to just providing information in the description field. Or we can document 'window' more thoroughly

PR #36, opened 2011-05-09
remove WorkerUtils and create a new "data storage" category

jackluo-zz:
1. remove WorkerUtils from header description in database js file since its openDatabase method is not supported in BB and PB
2. combine "web storage" and "database" categories into "data storage" category

### DIFF
--- a/api/html5_database.js
+++ b/api/html5_database.js
@@ -15,7 +15,7 @@
 */
 
 /**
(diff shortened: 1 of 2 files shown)


jeffheifetz:
You might want to talk to Marek about the "constructedBy" tag. It should help out with the window.openDatabase constructor.

PR #42, opened 2011-05-09
Various Fixes

marek:
[ ] accessor updates
@link fixes
@PB10 fixes

### DIFF
--- a/api/html5_media.js
+++ b/api/html5_media.js
@@ -648,8 +648,8 @@ TextTrackCue.prototype.getCueAsHTML = function() { };
 
 
 /**
-* A {TextTrackCueList} object represents a dynamically updating list 
-* of {TextTrackCue}s in a given order.
+* A {@link TextTrackCueList} object represents a dynamically updating list 
+* of {@link TextTrackCue}s in a given order.
 * @class
 * @BB60+
 * @PB10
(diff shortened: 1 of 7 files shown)


api/html5_media.js
```diff
@@ -902,19 +916,19 @@ MutableTextTrack.prototype.removeCue(cue) { }
...
 HTMLMediaElement.prototype.textTracks = { };
 
 /**
-* Creates and returns a new {MutableTextTrack} object, which is also 
+* Creates and returns a new {@link MutableTextTrack} object, which is also 
 * added to the media element's list of text tracks.
 * @param {String} kind kind is one of the following strings "subtitles", "captions", "descriptions", "chapters", "metadata"
 * @param {String} label text track label
 * @param {String} language language of the text track
 * @returns MutableTextTrack
```
jeffheifetz: If you put this type in braces it will create a dynamic link. You can also put a description
@returns {MutableTextTrack} Which is described thusly

jeffheifetz:
After all that work, do you think we should actually support a PB10+ ? I mean otherwise someone will have to do a find/replace when we start I guess

marek:
jheifetz: Not yet. Maybe best not to think that far ahead.

api/html5_media.js
marek: Thanks i missed one

PR #46, opened 2011-05-10
fixes for blackberry_ui_dialog

masamo:


cpinzon:
This looks good! we should have customAskAsync and standardAskAsync for PB10, and customAsk and standardAsk for BB50+

Can you cross reference the playbook stuff with this = http://www.blackberry.com/developers/docs/webworks/api/playbook/

jeffheifetz:
I believe the D_constants should be PB only and not BB, double check as per Carolina's comments

masamo:
According to http://www.blackberry.com/developers/docs/widgetapi/ , it seems that the D_constants are supported by BB.

jeffheifetz:
OK then the D_ constants are both

cpinzon:
She's right, it's the C_constants that do not exist in the PB. 

Can you please make sure that the other constants from the PB WebWorks (SIZE_constants, etc) are in there as well?

cpinzon:
const String  BOTTOM  = bottomCenter  
const String  CENTER  = middleCenter  
const String  TOP  = topCenter

masamo:
Only C_ and D_ constants are there on BB.

So I need to make another fix to make D_ on both.

cpinzon:
Yup: 
C_ constants - BB only
SIZE_ and others - PB only
D_constants - BB and PB

PR #55, opened 2011-05-10
Cleanup,  inherited methods/properties, Bugfix(s)

### LINKED ISSUE #41: 3 FilterExpressions
We have 3 FilterExpression objects now 

One in:
https://github.com/blackberry/WebWorks-API-Docs/blob/master/api/blackberry_find_FilterExpression.js
'blackberry.find.FilterExpression'

Two in:
https://github.com/blackberry/WebWorks-API-Docs/blob/master/api/blackberry_phone_Find_FilterExpression.js
Both are 'blackberry.phone.Find.FilterExpression'

Isn't the correct namespace for the constructor blackberry.find.FilterExpression ?

### LINKED ISSUE #52: @example when there is no default constructor
if a class is marked as a @namespace, or if a @constructedBy is used, the accompanying @example tags are not shown. Examples are tied to the default constructor. Maybe it should have its own section?

This does not affect examples given to members

marek:
6df9706:
Bugfix. removed Duplicate FindFilterExpression. Closes #41.

fecbe57:
inherited methods/properties

38944f6:
Cleanup: tweak to allow return type lists to wrap

22f9cb2:
Bugfix. Closes #52.


jeffheifetz:
Thanks for cleaning this up, great catch

PR #56, opened 2011-05-10
Changes to system event namespace, removed onsystemevent from TOC

cpinzon:
Removed onSystemEvent documentation from System in table of contents and it's reference in bb.system.events as an interface.


api/blackberry_system_event.js
```diff
@@ -126,23 +125,10 @@ blackberry.system.event.prototype.onHardwareKey = function(key,callback) { };
 /**
 * Assigns a listener for when the coverage status changes. 
 * @param {OnSystemEvent} callback Function to be called when coverage changes.  Only one function can be assigned to this event. To unregister the callback, call the onCoverageChange method and pass in null for the callback parameter.
```
jeffheifetz, line 127: These parameters are still referring to the interface you removed

PR #57, opened 2011-05-11
update desc and fix learn tag

tlai:


jeffheifetz:
Geolocation properties are all static when many shopuld likely be instance level. You can do this by prepending with ".prototype". There is an example in the cookbook if you need help

jeffheifetz:
Position.timestamp is still DOMTIMESTAMP instead of just Date

jeffheifetz:
PositionOptions has properties of type long instead of Number

PR #70, opened 2011-05-12
put learns tag for all page

tlai:
Duplicated all learns tag into all page of GPS

### DIFF
--- a/api/html5_geolocation.js
+++ b/api/html5_geolocation.js
@@ -130,7 +130,7 @@ Position = {
 PositionError  = {
 
     /**
-     * @description Error code. One of the PERMISSION_DENIED, POSITION_UNAVAILABLE or TIMEOUT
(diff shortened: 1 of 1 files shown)


api/html5_geolocation.js
```diff
@@ -58,220 +59,229 @@ Geolocation = {
...
-	 * @description Error code. One of the PERMISSION_DENIED, POSITION_UNAVAILABLE or TIMEOUT 
-	 * @BB50+
-	 * @PB10	
-	 * @readonly	 
-	 * @type Short
-	 */
+PositionError  = {
+
+    /**
+     * @description Error code. One of the PERMISSION_DENIED, POSITION_UNAVAILABLE or TIMEOUT
```
jeffheifetz: Could you make these link tags?

jeffheifetz:
How come it looks like a huge diff when most lines look the same?

tlai:
i also removed trailing space and convert TAB to Space.

tlai:
Good idea. Done

jeffheifetz:
I guess we have to switch to spaces eventually, so might as well start, its just annoying we lost all tracking. Looks good to me

PR #72, opened 2011-05-12
Horizontal scrollbar should now not show on the first display

cpinzon:


cpinzon:
Edit: there's a bug on the HTML5 RotationRate that I'm seeing, where the text is still wrapped and the image for the alpha property appears twice. no idea what's causing that. Please let me know if you can reproduce this and/or if you have a solution.

jeffheifetz:
This is because RotationRate has 2 description tags. Remove the duplicate and the issue is resolved.

Additionally DeviceOrientationEVent seems to have disappeared

cpinzon:
Ah, I missed that.

And yes, DeviceOrientationEvent was removed intentionally.

PR #80, opened 2011-05-15
Increasing support for function overloading

jeffheifetz:
Modified @constructedBy to mark a field/function as a constructor -- fields will not have parameters or brackets
Renamed some classes to their actual name and used traditional constructors
Added an example to the cookbook

### DIFF
--- a/api/html5_app_cache.js
+++ b/api/html5_app_cache.js
@@ -18,20 +18,40 @@
  * @class The applicationCache object is your programmatic access to the browser's app cache.
  * @notice {Warning (BlackBerry Only):}
  * Application Cache is only supported since OS version 6.0.0.418
- * @constructedBy {window.applicationCache} {Returns the ApplicationCache object that applies to the active document of that Window.}
- * var appCache = window.applicationCache
- * @constructedBy {worker.applicationCache} {Returns the ApplicationCache object that applies to the current shared {@link WebWorkers#applicationCache}.}
- * &lt;script type="text/javascript"&gt;
- *   var worker = new Worker('doWork.js');
- * &lt;/script&gt;
- *
- *doWork.js (the worker):
- *
- *   var appCache = self.applicationCache()
  * @toc {Cache} HTML5 ApplicationCache
  */
 
 ApplicationCache ={
+        
+        /**
+         * @constructedBy window.applicationCache 
+         * @field
+         * @description Returns the ApplicationCache object that applies to the active document of that Window.
+         * @example 
+         * &lt;script type="text/javascript"&gt;
+         *   var appCache = window.applicationCache;
+         * &lt;/script&gt;
+         * @BB60+
+         * @PB10
+         */
+        windowConstructor : undefined,
+        
+        /**
+         * @constructedBy worker.applicationCache 
+         * @field
+         * @description Returns the ApplicationCache object that applies to the current shared {@link Worker#applicationCache}.
+         * @example 
+         * &lt;script type="text/javascript"&gt;
+         *   var worker = new Worker('doWork.js');
+         * &lt;/script&gt;
+         *
+         *doWork.js (the worker):
+         *
+         *   var appCache = self.applicationCache()
+         * @BB60+
+         * @PB10
+         */
+        workerConstructor : undefined,
 
 		/**
 		* @constant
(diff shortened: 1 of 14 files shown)


api/html5_app_cache.js
```diff
@@ -18,20 +18,40 @@
...
+         *   var worker = new Worker('doWork.js');
+         * &lt;/script&gt;
+         *
+         *doWork.js (the worker):
+         *
+         *   var appCache = self.applicationCache()
+         * @BB60+
+         * @PB10
+         */
+        workerConstructor : undefined,
```
jeffheifetz, line 54: I wonder if we even need this since its documented in the worker class.

api/html5_database.js
```diff
@@ -15,19 +15,33 @@
...
+         * @param {String} displayName The display name of the database to be created
+         * @param {Number} estimatedSize The estimated size in bytes of the database
+         * @callback {function} creationCallback The callback will be invoked on database creation
+         * @callback {Database} creationCallback.database The newly created database
+         * @example
+         * Database db=Window.openDatabase('documents', '1.0', 'Offline document storage', 5*1024*1024, null);
+         * @PB10
+         * @BB50+
+         */
+        openDatabase : undefiuined,
```
jeffheifetz, line 43: Missing a @returns statement, will update

api/html5_media_source.js
```diff
@@ -42,6 +38,17 @@
 HTMLSourceElement = function() { };
 
 /**
+ * @field
```
cpinzon, line 41: this is a tag we've never use in our docs before. thoughts?

build/BBTemplate/class.tmpl
```diff
@@ -278,21 +272,19 @@
 	</if>
 
 <!-- ============================== Constructor Summary ========================= -->
-	<if test="(!data.isBuiltin() && data.is('CONSTRUCTOR')) || data.constructedBy">
+	<if test="(defined(constructors) && constructors.length)">
```
cpinzon, line 275: where is the defined() method established, and how does it evaluate if constructors exist?

build/BBTemplate/class.tmpl
```diff
@@ -278,21 +272,19 @@
 	</if>
 
 <!-- ============================== Constructor Summary ========================= -->
-	<if test="(!data.isBuiltin() && data.is('CONSTRUCTOR')) || data.constructedBy">
```
cpinzon, line 281: what happens to constructedBy? where does it get taken into account?

api/html5_media_source.js
jeffheifetz, line 41: The @field tag simply tells the parser to ignore whatever it sees and mark this as a field. It is the counterpart to @function.

In this case I am using it to denote that the fake constructor should not have ()'s put afterwards. This was necessary because we are marking all ways to get the object as constructors, even though not all necessarily are. IE window.applicationCache which retrieves an instance but does not construct a new one. Similarly document.createElement("source") must have the one exact parameter so to prevent the redrawing of brackets I marked it a field.

build/BBTemplate/class.tmpl
jeffheifetz, line 275: The defined method is used throughout and as the name implies it simply checks if the variable has been defined. I am not 100% sure where it is defined to be honest.
In truth I believe that the variable is guaranteed to be defined but the extra check can't hurt.

build/BBTemplate/class.tmpl
```diff
@@ -173,22 +173,27 @@
...
 	<if test="data.events.length">
 		{! var ownEvents = data.events.filter(function($){return $.memberOf == data.alias  && !$.isNamespace}).sort(makeNameSort()); !}
 		{! allChildren = allChildren.concat(ownEvents);!}
-	</if>	
-
+	</if>
+    {! var generateDefaultConstructor = !data.isBuiltin() && data.is('CONSTRUCTOR') && !data.noConstructor; !}
+    <if test="generateDefaultConstructor">
+        {! constructors.unshift(data); !}
+    </if>
```
jeffheifetz, line 196: The constructors variable is compiled up here along with all the collections we parse through the class template.  In lines above here we add the constructor fields and functions marked by @constructedBy to the constructors array.

PR #81, opened 2011-05-16
some updates to examples and added valid range of values for volumeLevel

gtanner:
Got rid of "new Array" in examples as no one uses that anymore in javascript and added range of values for volumeLevel


jeffheifetz:
Thanks for the help

PR #82, opened 2011-05-16
2D Context & Focus fix

marek:


jeffheifetz:
I think you should put a link in Canvas to the 2D Context class

api/html5_canvas_context2d.js
```diff
@@ -0,0 +1,934 @@
...
+*/
+CanvasRenderingContext2D.prototype.restore = function() { };
+
+
+/**
+* Add the scaling transformation described by the arguments to the 
+* transformation matrix. The x argument represents the scale factor 
+* in the horizontal direction and the y argument represents the scale factor in the vertical direction. The factors are multiples.
+* @param {Number} x
+* @param {Number} y
```
jeffheifetz: I think you should move the parameter descriptions from the method description to the parameters themselves ie

*@description Add the scaling transformation described by the arguments to the transformation matrix.
- @param {Number} x The scale factor in the horizontal direction.
- @param {Number} y The scale factor in the vertical direction.

What do you think?

marek: For the majority of the descriptions, it talks about the parameters in one big picture to describe the process. I don't feel as if picking through everything or duplicating the same info for each param, then rewording each general description is worth it. Is this just so we don't have a blank param description column?

marek:
Added a @link to the Canvas class.

api/html5_canvas_context2d.js
jeffheifetz: Mainly, yes. I just feel like it'll look bizarre if for just this one set of classes we have no parameter descriptions compared to all of the other ones.

marek: Perhaps not generating the third column if there are no descriptions? 
I find these types of descriptions worse than an empty column:

---

Method must multiply the current transformation matrix with the 
matrix described by:

<table>
<tr><td>m11</td><td>m21</td><td>dx</td></tr>
<tr><td>m12</td><td>m22</td><td>dy</td></tr>
<tr><td>0</td><td>0</td><td>1</td></tr>
</table>

@param {Number} m11  Is the (1,1) parameter of of the transformation matrix
@param {Number} m12  Is the (1,2) parameter of of the transformation matrix
@param {Number} m21  Is the (2,1) parameter of of the transformation matrix
@param {Number} m22  Is the (2,2) parameter of of the transformation matrix
@param {Number} dx  Is the x-scalar of the transformation matrix
@param {Number} dy  Is the y-scalar  of the transformation matrix
@BB60+
@PB10

---

Method takes six arguments, the first three representing the start 
circle with origin (x0, y0) and radius r0, and the last three 
representing the end circle with origin (x1, y1) and radius r1. 
...bla bla
@param {Number} x0 x-coord of start cicle
@param {Number} y0 y-coord of start circle
@param {Number} r0 radius of start circle
@param {Number} x1 x-coord of end circle
@param {Number} y1 y-coord of end circle
@param {Number} r1 radius of end circle
@BB60+
@PB10

PR #83, opened 2011-05-16
Fixed MKS1276534:Correct the description of blackberry.io.dir.listFiles funciton

qingbiao:


api/blackberry_io_dir.js
```diff
@@ -37,7 +37,7 @@
 blackberry.io.dir = { };
 
 /**
-* List all the files and sub-directories that reside in a given directory. 
+* List all the files that reside in a given directory. 
 * @param {String} path Path location of the directory.
 * @returns {String[]}
```
jeffheifetz, line 42: Could we add a return message as well ie
@returns {String[]} A string array containing all the files in the given directory.

I know its annoying repetition but it beats empty boxes

PR #95, opened 2011-05-17
Delta

ababut:
Add URI calls and examples for blackberry.app, blackberry.system and blackberry.identity.PIN.


api/blackberry_app.js
```diff
@@ -174,5 +178,65 @@ blackberry.app ={
...
+	/**
+	 * @uri 
+	 * @PB10
+	 * @throws {200} Success
+	 * @throws {400} BadRequest If the incorrect arguments are given
+	 * @throws {403} Forbidden Trying to access a feature that is not in the whitelist for your current domain
+	 * @throws {404} NotFound
+	 * @throws {500} ServerError Server is not working
+	 * @throws {501} NotImplemented Request made to a valid name space but an incorrect method name
+	 * @throws {503} ServiceUnavailable
```
jeffheifetz, line 192: I believe we are no longer placing these error codes here.

api/blackberry_identity.js
```diff
@@ -15,6 +15,11 @@
 */
 
 
+/**
+* @namespace
+*/
+blackberry = { };
```
jeffheifetz, line 21: The blackberry namespace is declared in blackberry.js (may need to wait for rebase from master)

api/blackberry_system.js
```diff
@@ -36,8 +36,8 @@ blackberry.system ={
 		
 		/**
 		 * @param {String} capability The capability being checked for.
-		 * @description This method will determine if the BlackBerry PlayBook or the BlackBerry Smartphone device is capable of the given service.<br>
-		 * <br> hasCapability() returns true if the BlackBerry device is capable of the service; false otherwise. <br>
+		 * @description This method will determine if the BlackBerry PlayBook is capable of the given service.<br>
+		 * <br> hasCapability() returns true if the BlackBerry PlayBook is capable of the service; false otherwise. <br>
```
jeffheifetz, line 40: Why change this if the method has BB50+ support?

api/blackberry_app.js
ababut, line 192: Should I take them out altogether or do some need to remain? The ones that get removed need to appear somewhere else?

----- Original Message -----

jeffheifetz, line 192: I believe we are only supposed to leave the application specific ones (if they still exist).

api/blackberry_system.js
ababut, line 40: Pure merge error, we probably had an older version that snuck through. Good catch.

api/blackberry_identity.js
ababut, line 21: Then it's fine, when master gets merged (and make absolutely sure that it's a merge, not a rebase!!), this problem will be fixed.

api/blackberry_app.js
ababut, line 192: Sorry about the reply junk, I decided to be fancy and reply via email. I'll remove the error codes. The app-specific error codes don't exist in this case iirc.

api/blackberry_system.js
jeffheifetz, line 40: There are others, check the diff and it should be obvious

ababut:
I just realized that I branched incorrectly. I will make a new branch and associated pull request that will have all the fixes coming from this review.

PR #98, opened 2011-05-17
Small api fixes

jeffheifetz:
Updated Geolocation to have proper callbacks and instance methods
Added HTML5 Shim notice to addditional classes
Re-worked Applicaton Cache Notice
Removed useless interface class


marek:
Looks good.

PR #106, opened 2011-05-17
Make the changes suggested in the review at GitHub

tohman:
Generated using the latest templates.

### DIFF
--- a/api/blackberry_media_camera.js
+++ b/api/blackberry_media_camera.js
@@ -1,0 +1,72 @@
+/*
+* Copyright 2010-2011 Research In Motion Limited.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


api/blackberry_media_camera.js
```diff
@@ -0,0 +1,72 @@
...
+* limitations under the License.
+*/
+
+/**
+ * @namespace
+ * <p>Using Camera API allows to invoke the native camera application to record video or take a picture and receive a path to the file stored.</p>
+ * <p>The methods are static and don't required to create an instance of a camera.</p>
+ * @toc {Media} Camera
+ * @BB50+
+ * @PB10
```
jeffheifetz: You can remove the support tags (@BB50) from the namespace level as they are ignored. These are automatically calculated based on all children (including inherited children) to prevent errors.

jeffheifetz:

PR #107, opened 2011-05-17
Updated the default return text for URI functions

jeffheifetz:
It now generates a default message and table describing the return data.

Here is an example

### DIFF
--- a/build/BBTemplate/class.tmpl
+++ b/build/BBTemplate/class.tmpl
@@ -1245,17 +1245,51 @@
 						</table>
 					</if>
 					
-					<if test="member.returns.length">
-						<br/>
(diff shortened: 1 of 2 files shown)


rwmtse:
Actually data is not null even if code=1, data always contains the params that were passed in, even if the call didn't finish successfully.

rwmtse:
And as discussed, if we are having that one document that covers all URI-based APIs in general, the table should be in that document instead of in every single URI-based API

jeffheifetz:
So maybe just a blurb saying that for full documentation checkout the page and a link?

rwmtse:
Yea, I think so.

PR #108, opened 2011-05-17
Add URI calls for app and system and the PB version of the PIN API.

ababut:


api/blackberry_app.js
```diff
@@ -175,4 +180,58 @@ blackberry.app ={
...
-};
\ No newline at end of file
+	
+	
+	/**
+	 * @uri 
+	 * @PB10
+	 * @returns {JSON of all the properties}
+	 * {
+	 * 	"data":{
```
jeffheifetz, line 190: Shouldn't you have code and msg here?

jeffheifetz, line 190: Based on pull #107 I guess you should ignore this

jeffheifetz:

ababut:
Thanks! I am proposing that we should only show the API-specific data in each API's docs. The rc/message parts are redundant, they can be documented in a common spot together with the HTTP return codes.

PR #109, opened 2011-05-19
Added permission tag requires use_camera permission.

tohman:


api/blackberry_media_camera.js
```diff
@@ -20,6 +20,7 @@
  * <p>The methods are static and don't required to create an instance of a camera.</p>
  * @toc {Media} Camera
  * @featureID blackberry.media.camera
+ * @permission use_camera This permission should be granted in order to be able to invoke native camera.
```
jeffheifetz: How about

```
@permission use_camera Permits the user to invoke the native camera application.
```

tohman: The change is coming..

api/blackberry_media_camera.js
```diff
@@ -20,6 +20,8 @@
  * <p>The methods are static and don't required to create an instance of a camera.</p>
  * @toc {Media} Camera
  * @featureID blackberry.media.camera
+ * @permission use_camera Permits the user to invoke the native camera application.  
+ * @permission [access_shared] Permits the user to retrieve a path to the file stored in the system.
```
jeffheifetz: I know I just bugged you about the permissions, but do these sound clearer. I no longer like user and I think shorter is better based on how they currently display.

What do you think?

```
@permission use_camera Permits your app to use the camera.  
@permission [access_shared] Permits your app to access the shared files and folders.
```

tohman: Like the first one.
Second permission is needed in the case user want to listen for the onCapture event and interested in its filePath parameter, for different reasons i.e. saving, storing, displaying, not necessarily to access the file itself.

jeffheifetz: Doesn't that just mean it permits read/write access to the shared folder...What you do with it is up to the developer, right?

tohman:
To retrieve the full path to file saved there is a need in action script to analyze shared folder and for that a permission required.
It shouldn't be optional since onCaptured to whom filePath is callback is passed is a required parameter.

jeffheifetz:

PR #111, opened 2011-05-19
bbPlugin permissions update

cpinzon:


cpinzon:
This was just a small addition of two lines to bbPlugin that nibble the name of the @permissions tag data and allow proper functionality. It was originally meant to be in the permissions commit but it wasn't updated.

jeffheifetz:
Can you double check that you are exclusively using spaces, it seems bizarre that the diff is so enormous

jeffheifetz:
Also, can you include the change in the class.tmpl that changes it from <permission> to rim:permit

cpinzon:
Yes I am. The formatting of the code wasn't indented so it was rather unreadable to me. I passed it through a tool called jsbeautifier and it cleaned it up. That's all it was. It should be good to pull into delta.

cpinzon:
Oh right. Will do that.

build/BBTemplate/class.tmpl
```diff
@@ -288,7 +288,7 @@
 			
 			<for each="item" in="optional">
 				<tr><td class="scriptTd">
-				- &lt;permission="{+item.name+}" /&gt; {+item.desc+}
+				- &lt;rim:permit="{+item.name+}" /&gt; {+item.desc+}
```
jeffheifetz: I believe the style is 
    &lt;rim:permit&gt; permission_name &lt;/rim:permit&gt;

cpinzon:
The permissions show now fully reflect what the needed configurations in a playbook webworks app are. This should be ready to close and pull into delta.

jeffheifetz:
I agree, this is ready to merge

PR #113, opened 2011-05-20
Collapsible toc

cpinzon:
Preliminary collapsible toc div. Uses jquery functions to dynamically collapse and expand the toc.

This is a very rudimentary version of this functionality, since the refreshing of the body once the toc vanishes is a big jagged.

I am looking for suggestions on where the toggling div should be placed, and what form should it take (a button?)


tneil:
I'll have to check it out... I have scrollable content working that I want to make sure is compatible with the collapsible code

jeffheifetz:
Also I think this should be targeted towards the master branch

PR #123, opened 2011-05-26
Cleaning up class temlate and adding body tag

jeffheifetz:
Added a proper body tag (previously was missing the <body> although the close was there)
Added a div around all of the content (With the scrollable class and content id)
Fixed up some mismatched and misused tags


jeffheifetz:
This pull request is ready to merge

PR #124, opened 2011-05-26
We now have a scrollable TOC and body contents

tneil:
Updated templates and content to allow for scrollable TOC and content body for browsers that are not IE.  IE was just to bloody crazy to make work :(

This also requires Jeff's pull request:
https://github.com/blackberry/WebWorks-API-Docs/pull/123

And we will have to adjust the show/hide TOC pull request from Carolina


jeffheifetz:
This pull request is ready to merge

PR #126, opened 2011-05-26
Payment API Revised and Corrected

tohman:


api/blackberry_payment.js
```diff
@@ -0,0 +1,181 @@
...
+     */
+    developmentMode: false
+
+};
+
+/**
+ * @class
+ * @toc {Payment} Purchase
+ */
+Purchase = {
```
jeffheifetz: If we have all of the arguments documented in the purchase method, I don't think we need this class at all.

tohman:
The reason this object was re-factored is because it passed for both 'getExistingPurchases' and 'purchase' as a parameter upon success (callbackOnSuccess.data), instead it appears only at one place and linked from the methods mentioned earlier.

api/blackberry_payment.js
jeffheifetz: I understand now why we need it. Should we make it blackberry.payment.puchase?

Also all of these attributes are instance level. Please change it to use .prototype. IE

```
Purchase.prototype.transactionID = null;
```

api/blackberry_payment.js
```diff
@@ -0,0 +1,179 @@
...
+     * &lt;/script&gt;
+     */
+    purchase : function (args, callbackOnSuccess, callbackOnFailure) {
+    },
+    /**
+     * @type Boolean
+     * @description Sets the development mode used in the application. If development mode is set to true, the application does not contact the Payment Service server for any transactions. For purchases, a simulated purchase screen is displayed, allowing the user to choose the result of the purchase. For retrieving existing purchases, only simulated successful purchases are returned. This mode is useful for testing how your application handles the possible results without requiring network connections or currency. THIS MODE SHOULD NOT BE USED IN PRODUCTION CODE. If development mode is set to false, purchases and retrievals of existing purchases proceed normally, contacting the Payment Service server as necessary. This is the default development mode, and applications in production should not modify it.
+     * @PB10
+     */
+    developmentMode: false
```
jeffheifetz, line 122: Does this have a default value?

api/blackberry_payment.js
tohman: I'm changing it to be define with prototype. 
Not sure if blackberry.payment. should appear in front of purchase because you can't create/access it in that way and it may mislead.

jeffheifetz:
This pull request is ready merge

PR #127, opened 2011-05-26
Cleaning up HTML in class template

jeffheifetz:
Cleaning up the class template
Added a body tag 
Added a div around the whole body
Fixed HTML with mismatched open and close tags


jeffheifetz:
There is a bug introduced in some of the linking, do not merge

PR #128, opened 2011-05-26
Cleaning up HTML in class template

jeffheifetz:
Updating the HTML to have a body and div tag surrounding the content as well as fixing mismatched tags


jeffheifetz:
This is ready to merge

PR #130, opened 2011-05-26
Fix onCoverageChange and onHardwareKey, they should be static functions, not instance functions

rwmtse:


jeffheifetz:
This is ready to merge

PR #131, opened 2011-05-26
Add new push API jsdoc

qingbiao:


api/blackberry_push.js
```diff
@@ -30,26 +50,16 @@ blackberry.push = {
...
-	* <br/><br/>
-	* At that point it will then pass the data to the subscribed callback function. To subscribe to the same previously 
-	* opened push port on a new web page, a secondary call to openPushListener is called. In this scenario it will 
-	* reassign the callback provided to the already open port. 
-	* <br/><br/>
 	* If calling openPushListener on a page where the port is already open, the callback will be updated to use the 
 	* latest one passed in. <b>NOTE:</b> In such cases where consecutive calls to openPushListener occurs on the page,
 	* the transport argument will only be used on the first call, and ignored for all others. Any calls to 
 	* openPushListener should have matching calls to closePushListener to ensure ports are not left open accidentally.
+	* @deprecated This API is deprecated, please use opebBESPushListener or openBISPushListener instead.
```
jeffheifetz: use @link tags so that a link will be generated

api/blackberry_push.js
```diff
@@ -89,12 +99,143 @@ blackberry.push = {
...
+	* &lt;/script&gt;
 	*/
-	closePushListener : function(port) { }
-
+	closePushListener : function(port) { },
 
+	/**
+	* Opens the push listener to allow an application to listen for BES pushes. Only one port can be opened at a time if openBESPushListener
+	* or openBISPushListener functions are used.
+	* Furthermore, the callback function must return an integer (one of constants described in blackberry.push.Data)
```
jeffheifetz: Then shouldn't there be an @returns tag like

```
@returns {Number} One of the constants defined in {@link blackberry.push.Data}
```

qingbiao: This is onData callback return; not openBESPushListener retur. How do i write this?

jeffheifetz: This is not possible and is captured in Issue #140

api/blackberry_push.js
```diff
@@ -30,26 +50,16 @@ blackberry.push = {
 	* <br/><br/>
 	* Furthermore, the callback function must return an integer (one of constants described in blackberry.push.Data)
```
jeffheifetz: Please use link tags here as well

api/blackberry_push.js
jeffheifetz: Also, please use @link tags

api/blackberry_push.js
```diff
@@ -88,13 +98,144 @@ blackberry.push = {
...
+	*     try {
+	*       var ops = { port : port, appId : appId, serverUrl : serverUrl, wakeUpPage : 'push.htm', maxQueueCap : max };
+    *       blackberry.push.openBISPushListener(ops, onData, onRegister, onSimChange);
+    *     }
+    *     catch (err) {
+    *       alert(err);
+    *     }     
+	*   } 
+	*
+	*   function onRegister(status) {
```
jeffheifetz, line 194: Do we have these status codes documented anywhere other than in the example?

PR #145, opened 2011-06-02
Fixed incorrect method signature

gtanner:
See:
https://github.com/blackberry/WebWorks/blob/master/api/utils/src/blackberry/utils/GenerateUniqueIdFunction.java#L26


jeffheifetz:
Great catch!
This change is ready to merge

PR #152, opened 2011-06-03
Updated info for supported components

tohman:


cdelcol:
This is good to go. Alert comment is part of an example and is ok.

PR #156, opened 2011-06-03
Changed type of port property to Number from String

tohman:


jeffheifetz:
This change is ready to merge

PR #159, opened 2011-06-07
Epsilon master merge

jeffheifetz:


ababut:
The change coming from acaa1f7 is showing in be67af1 and 77ecbe0. Also the template changes appear duplicated in 77ecbe0 and 695c234.

ababut:
I would recommend to merge branches with the --no-commit option so that you can give them non-default names. It's very hard to read at a glance what's in each merge. You have to click on it in the list above to get a complete commit message.

PR #161, opened 2011-06-13
fix for blackberry.system.hasPermission method

gtanner:
hasPermissions here:
https://github.com/blackberry/WebWorks/blob/master/api/system/src/blackberry/system/SystemNamespace.java

hasPermission here:
https://github.com/blackberry/WebWorks/blob/master/api/system/src/blackberry/system/SystemNamespace.java#L40

I would assume that this is a bug in the docs.


cdelcol:
Correct. Good to go.

Issue #122, opened 2011-05-25
There is no clear use of the @event tag


jeffheifetz:
Currently we do not have a consistent system for deciding when a function should be marked as an event.

HTML5 Element Touch Events - have all of its functions marked as events

System EVents - has its very similar functions just as functions

I believe we need to come up with a more consistent use of the @event tag or remove it.

marek:
I thought this was the difference:

@event is where the event member is triggered directly.

/**
- @event
- @param {Event} event
  */
  Element.onclick = function(event) // Some event will call onclick().

@callback, would be where something triggers the callback passed to the method member as a parameter. 
I see system events using @callback. 

/**
- @function
- @param {String} event event name
- @callback {Function} callback callback to invoke on event
  */
  element.addEventListener = function (event, callback) { }

example:
function myCallback { }
element.addEventListener('click',myCallback )

Seems to be used correctly in both Touch event and system events?

jeffheifetz:
The major distinction I see in your example is not really how it is triggered but how the value is assigned. This seems reasonable to me, perhaps something to keep in mind when developing new APIs

```
@event
class.onEvent = function
```

vs

```
@callback
class.onEvent(function)
```

PR #173, opened 2011-06-17
BBM api addition docs

rburbidge:
First submit of BBM Platform docs.


api/blackberry_bbm.js
```diff
@@ -0,0 +1,22 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * @namespace Provides access to BlackBerry Messenger services.
+ */
+blackberry.bbm = {
+
+};
\ No newline at end of file
```
jeffheifetz: This file adds no new functionality and will create an empty page. I think it should be removed

rburbidge: How can I link to the blackberry namespace if I remove this?

jeffheifetz:
I really like the tables you've put in to show the possible values, however this is something unlike any of our other docs. I'm not sure what people think whether this is something we should adopt and add this to other APIs, reject it, or accept inconsistency.

Another issue is that their style is so different from all of our own HTML

An example can be seen below

api/blackberry_bbm.js
jeffheifetz: I'm not sure what you mean by linking here, but its not like we enforce the docs to be valid Javascript

api/blackberry_bbm_platform_ui_menu.js
```diff
@@ -0,0 +1,100 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * @toc {BBM} Menu
+ * @featureID blackberry.bbm.platform
+ * @namespace Provides the ability to add menu items within BBM. Menu items persist across device resets.
+ * @BB50+
+ */
+blackberry.bbm.platform.ui.menu = {
```
jeffheifetz: Don't you need some way to get the menu items you've already created? Like getMenuItems?

jeffheifetz:
Weren't Session and Channel supposed to have a common method for closing/ending for consistency sake?

api/blackberry_bbm_platform_io_Connection.js
```diff
@@ -0,0 +1,288 @@
...
+     * Disables hosting on this connection.
+     * @BB50+
+     */
+    this.disableHosting = function() { };
+    
+    /**
+     * <code>true</code> if the user is hosting on this connection; <code>false</code> otherwise.
+     * @BB50+
+     */
+    this.hosting = false;
```
jeffheifetz: needs an @type Boolean tag

jeffheifetz:
Since the creation of the propertyCB item we introduced the use of the @event tag and there is currently no use case for @propertyCB. Please convert all of them to @event tags.

For more information see Issue #122

api/blackberry_bbm_platform_io_IncomingJoinRequest.js
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * @featureID blackberry.bbm.platform
+ * @class A request to join received on the host's side.
+ * @extends blackberry.bbm.platform.io.JoinRequest
+ * @BB50+
+ */
+blackberry.bbm.platform.io.IncomingJoinRequest = function() {
```
jeffheifetz, line 24: Is there a reason there is no TOC entry?

api/blackberry_bbm_platform_io_OutgoingJoinRequest.js
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * @featureID blackberry.bbm.platform
+ * @class A request to join sent from the peer's side.
+ * @extends blackberry.bbm.platform.io.JoinRequest
+ * @BB50+
+ */
+blackberry.bbm.platform.io.OutgoingJoinRequest = function() {
```
jeffheifetz, line 24: Same question about TOC entry

api/blackberry_bbm_platform_io_Session.js
```diff
@@ -0,0 +1,92 @@
...
+    // Session methods //
+    /////////////////////
+    
+    /**
+     * Ends the connection for all participants.
+     * <p>All participants will be removed from {@link joinedUsers}. All pending invitations will be
+     * removed and {@link pendingUsers} will be set to <code>0</code>.
+     * <p>{@link onSessionEnded} will be called for all users currently in the connection.
+     * <p>The connection will become inactive for the current user, and should not be used afterwards.
+     * @throws IllegalStateException if the current user has already ended the connection.
```
jeffheifetz: Should be 

```
@throws {IllegalStateException} if the current user has already ended the connection.
```

api/blackberry_bbm_platform_io_Session.js
```diff
@@ -0,0 +1,92 @@
...
+     * @BB50+
+     */
+    this.end = function() { };
+    
+    /**
+     * Removes the current user from the connection, but leaves it active for all other participants.
+     * <p>{@link onUserLeft} will be called for all other users currently in the connection.
+     * <p>The connection will become inactive for the current user, and should not be used afterwards.
+     * @throws IllegalStateException if the current user has already left the connection.
```
jeffheifetz: Should be 

```
@throws {IllegalStateException} if the current user has already left the connection.
```

api/blackberry_bbm_platform_self.js
```diff
@@ -0,0 +1,92 @@
...
+         * @BB50+
+         */
+        setPersonalMessage: function(personalMessage, onComplete) { },
+        
+        /**
+         * @description Sets the user's status. A dialog will be presented to the user to allow
+         * or deny the change.
+         * @param {String} status The status: one of <code>"available"</code> or <code>"busy"</code>.
+         * @param {String} message The optional status message. If <code>null</code> then the default message of
+         * either "Available" or "Busy" will be used.
```
jeffheifetz, line 75: The two parameters disagree on the casing of the possible values, is this on purpose? Also no default value is provided on the first parameter, what happens with incorrect values?

jeffheifetz:
All of your namespace/classes need to have a beta tag addeed. It would look something like this

```
@beta For more information about the BBM API Beta ...
```

api/blackberry_bbm_platform_self_profilebox.js
```diff
@@ -0,0 +1,192 @@
...
+     *     
+     *     if(! blackberry.bbm.platform.self.profilebox.hasIcon(iconId)) {
+     *         blackberry.bbm.platform.self.profilebox.registerIcon(iconId, iconInfo.uri);
+     *     }
+     * }
+     * &lt;/script&gt;
+     * 
+     * @BB50+
+     */
+    registerIcon : function(iconID, iconURI) {
```
jeffheifetz: What if the iconID already has an icon registered to it? I assume it overwrites, but this isn't documented

api/blackberry_bbm_platform_self_profilebox.js
```diff
@@ -0,0 +1,192 @@
...
+blackberry.bbm.platform.self.profilebox = {
+
+    /**
+     * <code>true</code> if the profile box is accessible and can be updated; <code>false</code> otherwise.
+     * The profile box is not accessible when the device is in mass storage mode.
+     * @type Boolean
+     * @readOnly
+     * @BB50+
+     */
+    accessible : false,
```
jeffheifetz, line 65: Throughout the class you say "@throws {UserProfileBoxAccessException} If the device is in mass storage mode." but I wonder if in future there may be other reasons added. Might it not make more sense to reference this property which will list all of the possible reasons?

api/blackberry_bbm_platform_self_profilebox_ProfileBoxItem.js
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @featureID blackberry.bbm.platform
+ * @class A profile box item. Consists of text and possibly an icon.
+ * @BB50+
+ */
+blackberry.bbm.platform.self.profilebox.ProfileBoxItem = function() {
```
jeffheifetz, line 23: Why is there no TOC entry?

api/blackberry_bbm_platform_ui.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+/**
+ * @featureID blackberry.bbm.platform
+ * @namespace Provides integration services with the BBM UI.
+ * @BB50+
+ */
+blackberry.bbm.platform.ui = {
+};
\ No newline at end of file
```
jeffheifetz: Another file to be removed

api/blackberry_bbm_platform_ui_menu.js
```diff
@@ -0,0 +1,100 @@
...
+     * the <code>event="BBMmenuinvite"</code> and <code>param=menuItemId</code>. If the application
+     * is not running then it will be launched. The invocation of
+     * {@link blackberry.bbm.platform.io.onConnectionCreated} will not occur until the callback is assigned.</p>
+     * @param {Object} options Options.
+     * @param {Number} options.menuItemId Unique menu item ID.
+     * @param {String} options.label The menu item label.
+     * @param {Number} options.order Display order of this menu item; lower order items will be displayed first.
+     * @param {String} options.invitationMessage The invitation message displayed to the recipient.
+     * @throws {IllegalArgumentException} if <code>options.label</code> is not provided or <code>options.label.length > MAX_LENGTH_MENU_ITEM_LABEL</code>.
+     * @throws {IllegalArgumentException} if <code>options.invitationMessage</code> is not provided or <code>options.invitationMessage.length > MAX_LENGTH_MENU_ITEM_MESSAGE</code>.
```
jeffheifetz: If label and invitationMessage aren't optional, shouldn't they be required parameters outside of the JSON  and just leave optional parameters inside?

api/blackberry_bbm_platform_users.js
```diff
@@ -0,0 +1,236 @@
...
+ */
+
+/**
+ * @toc {BBM Users} Other Users
+ * @featureID blackberry.bbm.platform
+ * @namespace Provides access to user information. The current user can be accessed from
+ * {@link blackberry.bbm.platform.self}.
+ * @BB50+
+ */
+blackberry.bbm.platform.users = {
```
jeffheifetz, line 60: This file seems to differentiate from the others and our conventions in terms of parameter order. The ordering should be:

```
1. Options Object
2. Non-optional arguments
3. Callbacks
4. Optional arguments
```

api/blackberry_bbm_platform_users_ContactInvitation.js
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @featureID blackberry.bbm.platform
+ * @class A BBM contact invitation to be used in {@link blackberry.bbm.platform.users.inviteToBBM}
+ * @param {String} PIN The user's PIN.
+ * @param {String} name The user's name.
+ * @constructor
+ * @BB50+
+ */
+blackberry.bbm.platform.users.ContactInvitation = function(PIN, name) {
```
jeffheifetz: Why is there no TOC entry?

api/blackberry_bbm_platform_users_ContactInvitation.js
```diff
@@ -0,0 +1,43 @@
...
+     */
+    this.PIN = "";
+
+    /**
+     * @description The user's name.
+     * @type String
+     * @readOnly
+     * @BB50+
+     */
+    this.name = "";
```
jeffheifetz: What does this represent? Is it the name I give to a contact on my list? Does it have any forced connection to the PIN?

ababut:
I like the idea of presenting the possible values of in some format, but I'm not a fan of these Javadoc-style tables. 

As a starting point we can use a collapsible shaded section similar to that used for the whitelist features. It can be closed by default to avoid the visual clutter. 

What do you guys think?

api/blackberry_bbm_platform_ui_menu.js
```diff
@@ -0,0 +1,100 @@
...
+    MAX_LENGTH_MENU_ITEM_MESSAGE : 0,
+        
+    /**
+     * @description Adds a invitation menu item to the BBM contact list screen. If a menu item with
+     * the same <code>menuItemId</code> already exists for the application then it will be overwritten.
+     * <p>When the menu item is invoked, an invitation will be sent to the selected contact. On the
+     * inviter's side, {@link blackberry.bbm.platform.io.onConnectionCreated} will be invoked with
+     * the <code>event="BBMmenuinvite"</code> and <code>param=menuItemId</code>. If the application
+     * is not running then it will be launched. The invocation of
+     * {@link blackberry.bbm.platform.io.onConnectionCreated} will not occur until the callback is assigned.</p>
```
ababut: The statement is confusing/redundant - if it's a callback, it can't be invoked unless assigned. If the behaviour of that event is different with no callback assigned, it needs to be described in more detail. 

I recommend removing it from addInvitationMenuItem because it's confusing.

api/blackberry_bbm_platform_ui_menu.js
```diff
@@ -0,0 +1,100 @@
...
+     * @throws {IllegalArgumentException} if <code>options.invitationMessage</code> is not provided or <code>options.invitationMessage.length > MAX_LENGTH_MENU_ITEM_MESSAGE</code>.
+     * @throws {IllegalArgumentException} if <code>MAX_MENU_ITEMS</code> has been reached.
+     * @BB50+
+     */
+    addInvitationMenuItem : function(options) {
+    
+    },
+
+    /**
+     * Returns <code>true</code> if the application has already added the menu item; <code>false</code> otherwise.
```
ababut: Is the menu item ID returned by addInvitationMenuItem?

api/blackberry_bbm_platform_users.js
```diff
@@ -0,0 +1,236 @@
...
+     * @readOnly
+     * @BB50+
+     */
+    contactsWithApp : [],
+
+    /**
+     * @description Shows Contact Picker dialog allowing the user to select contacts.
+     * @param {Object} options Object containing contact picker options.
+     * @param {String} [options.title] Title of the contact picker dialog.
+     * @param {String} [options.type] Type of users to include in the dialog. May be "contactswithapp".
```
ababut, line 74: What does "contactswithapp" do? Needs to get documented.

api/blackberry_bbm_platform_users.js
```diff
@@ -0,0 +1,236 @@
...
+     * @param {blackberry.bbm.platform.users.BBMPlatformUser[]} [options.users] Users shown in the
```
ababut, line 75: Does users have a default value/behaviour? If not, it should be reflected in the code example.

api/blackberry_bbm_platform_users.js
jeffheifetz, line 74: I believe its a static readable property that gives you all of the contacts with the app installed. These can then be used for invites or other actions

jeffheifetz:
I like the idea of collapsing but it would require template changes so they are always consistent.  I'm thinking something like @value {value} description.

However it does not address the issue of consistency unless we take on the extra work of adding this to the existing docs.

api/blackberry_bbm_platform_io_IncomingJoinRequest.js
rburbidge, line 24: This is a small part of the hosted connections framework so I don't see too much merit.

I would have put a TOC entry for the two "add user" frameworks but it appears you can only have entries for whole objects. So I only have an entry for IO, Channel, and Session, hoping to lead them to the IO object.

api/blackberry_bbm_platform_self.js
rburbidge, line 75: message is supposed to be optional. If they do not provide it then it defaults to "Available" or "Busy" based on the value of status.

api/blackberry_bbm_platform_self_profilebox.js
rburbidge: It would overwrite it, but in practice they should never be overwriting an icon. I only allow it in case the developer wants to change around their icon-ID maps while in development.

api/blackberry_bbm_platform_self_profilebox_ProfileBoxItem.js
rburbidge, line 23: There's a TOC for self.profilebox. I don't think that mentioning this object in TOC adds value.

api/blackberry_bbm_platform_ui_menu.js
rburbidge: We don't have it in Java right now. We plan to add it after RTM.

api/blackberry_bbm_platform_ui_menu.js
rburbidge: Agreed. I will remove it.

api/blackberry_bbm_platform_ui_menu.js
rburbidge: All of the JSON parameters here are required though. I was recommended to use JSON for this method.

api/blackberry_bbm_platform_ui_menu.js
rburbidge: No the developer will set the menu item ID in the options parameter.

api/blackberry_bbm_platform_users_ContactInvitation.js
rburbidge: There are a few ways to invite users to your contact list. If possible I'd like to have them all under a single TOC entry. But they are split between objects so I'll need the ability to have TOC entries not based on objects.

api/blackberry_bbm_platform_users_ContactInvitation.js
rburbidge: It's the display name in BBM which is independent of the PIN.

api/blackberry_bbm_platform_users.js
rburbidge, line 60: Can you point out the methods with issues?

api/blackberry_bbm_platform.js
```diff
@@ -0,0 +1,175 @@
...
+ */
+
+/**
+ * @toc {BBM} BBM Platform
+ * @featureID blackberry.bbm.platform
+ * @namespace Provides access to the BBM Social Platform.
+ * 
+ * <h1>Required BBM Version</h1>
+ * 
+ * BBM Social Platform APIs come with BBM6 and later.
```
ababut: You mention that the platform APIs come with BBM6 and later, but the supported platforms say 5.0+.

api/blackberry_bbm_platform.js
```diff
@@ -0,0 +1,175 @@
...
+     * @description Registers for access to BBM Platform.
+     * <p>The application must assign a callback to {@link blackberry.bbm.platform.onAccessChanged}
+     * before registering. During registration, a dialog will be shown to guide the user through the registration process.
+     * The application should wait until <code>platform.onAccessChanged</code> is invoked before continuing.</p>
+     * <h4>Application in Test Environment</h4>
+     * Applications must provide a UUID used to identify the application in the test
+     * environment. If the application is in App World then the UUID will not be used. The same
+     * UUID should be used for future releases of the same application; otherwise communication
+     * between them will not be possible. The UUID must be a randomly generated 36-character UUID.
+     * Any UUID generator can be used.
```
ababut, line 56: Can we generate this UUID for them under the covers in Java? 

In general, we want to avoid asking JS developers to work with and manage IDs of any kind, unless absolutely necessary.

api/blackberry_bbm_platform.js
```diff
@@ -0,0 +1,175 @@
...
+     *         // Access blocked by user
+     *     } else if (status == "rim") {
+     *         // Access blocked by RIM
+     *     }
+     *     // Listen for other status...
+     * };
+     * 
+     * // Register with the platform
+     * blackberry.bbm.platform.register({
+     *     uuid: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" // Randomly generated UUID
```
ababut, line 78: Until we finalize this UUID, let's encourage them to use the WW method for generating one:

Check out Utils.generateUniqueId.

api/blackberry_bbm_platform.js
```diff
@@ -0,0 +1,175 @@
...
+     * @param {blackberry.bbm.platform.self.profilebox.ProfileBoxItem} param The parameter associated with <code>reason</code>.
+     * @example
+     * &lt;script type="text/javascript"&gt;
+     * 
+     * // Trigger an action in the application when a profile box item is selected
+     * blackberry.bbm.platform.onAppInvoked = function(reason, param) {
+     *     if(reason == "profilebox") {
+     *         var boxItem = param;
+     *         var achievementID = boxItem.cookie;
+     *         challenge(achievementID);
```
ababut: What does the challenge(acheievementID) method do? We should show it in the example to not confuse the copy-paste-pray masses :)

api/blackberry_bbm_platform_ui_menu.js
```diff
@@ -0,0 +1,100 @@
...
+     * @description The maximum length of a menu item invitation message.
+     * @type Number
+     * @constant
+     * @default 128
+     * @BB50+
+     */
+    MAX_LENGTH_MENU_ITEM_MESSAGE : 0,
+        
+    /**
+     * @description Adds a invitation menu item to the BBM contact list screen. If a menu item with
```
ababut: typo: "an invitation"

api/blackberry_bbm_platform_ui_menu.js
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright 2010 Research In Motion Limited.
```
ababut: The MenuItem API has a dangerous pitfall for WebWorks developers: they have to manage persistent IDs across application restarts and page switches. The trick is that BBM persists their menu entries, but unless they get a persistence mechanism in place, WW developers won't be able to track their menu items. Many simpler WW apps won't use any form of persistence as it adds a significant amount of complexity.

Also, IDs should be managed by the API for the reason above. I see a developer starting his app, incrementally using hasMenuItem to check integers from 0 to the first unused one and adding his item there. Then he has to persist that ID to use his menu item on the next page/app restart. 

This sort of low-level management should be handled by the extension. In the future, if BBM SDK's MenuItemManager starts managing it, you can't change the signature and pass the benefit downstream.

ababut: I don't think that this is a showstopper for the Beta release, but I'm posting it so we have it on the backlog.

api/blackberry_bbm_platform_users.js
ababut, line 74: Gotcha, I see it now!

api/blackberry_bbm_platform_users_BBMPlatformUser.js
```diff
@@ -0,0 +1,96 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @featureID blackberry.bbm.platform
+ * @class A platform user.
```
ababut: We need to be a little more descriptive here :) 

If its a base class with informational fields that others inherit from, let's call that out. Can it be instantiated by itself?

api/blackberry_bbm_platform_users_BBMPlatformUser.js
```diff
@@ -0,0 +1,96 @@
...
+    this.statusMessage = "";
+    
+    /**
+     * @description The user's unique ID. This will be <b>different</b> for multiple instances of the
+     * user (BlackBerry, PlayBook, etc.)
+     * @type String
+     * @readOnly
+     * @BB50+
+     */
+    this.handle = 0;
```
ababut, line 89: Handle and PPID have the same description which is very confusing. I'm guessing that the handle is used to identify the device and the PPID the user?

If that's the case, we need to be specific. However, since the API doesn't support PlayBook, and I don't think that these values are used as parameters to any calls (I may have missed it though), do we need them?

api/blackberry_bbm_platform_users_BBMPlatformUser.js
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * @description The personal message.
+     * @type String
+     * @readOnly
+     * @BB50+
+     */
+    this.personalMessage = "";
+
+    /**
+     * @description The status. One of <code>"available"</code> or <code>"busy"</code>.
```
ababut: I would make the description just: One of <code>"available"</code> or <code>"busy"</code>.

api/blackberry_bbm_platform_users_BBMPlatformUser.js
```diff
@@ -0,0 +1,96 @@
...
+     * @type String
+     * @readOnly
+     * @BB50+
+     */
+    this.status = "";
+
+    /**
+     * @description The status message.
```
ababut, line 76: I would make this description something like: The status message set by the user in BBM.

Basically to be clear that we're reading his status msg.

api/blackberry_bbm_platform_ui_menu.js
rburbidge: There is no reason to persist the menu items IDs. The developer should not be adding menu items dynamically.

The must have a statically defined set of menu items (i.e. the same app on different devices should add the same menu items under the same IDs). This is so that the invitation can be correctly interpreted on the other side.

For example, all instances of a games app would define the following items regardless of device:

ID 1: checkers
ID 2: chess
ID 3: backgammon

If another app is assigning the wrong IDs then a "chess" invite might show up as a "checkers" invite on the other side.

A getMenuItems() API is planned but will not be available in this release. I think the API is fully functional as-is.

api/blackberry_bbm_platform_users.js
rburbidge, line 75: For this API they should define either users, type, or both. Not sure how is best to convey that in the docs.

api/blackberry_bbm_platform_users_BBMPlatformUser.js
rburbidge: Good point. No you can't instantiate it yourself. I can add some information about where users are obtained from/how they are used.

api/blackberry_bbm_platform_users_BBMPlatformUser.js
rburbidge, line 76: Hmm, all of them are things that can be set in BBM or in third party apps. How about I add more information in the object level docs.

api/blackberry_bbm_platform_users_BBMPlatformUser.js
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * @description The display picture, encoded as a base64 image string.
+     * @type String
+     * @readOnly
+     * @BB50+
+     * @example
+     * &lt;script type="text/javascript"&gt;
+     * // Show a contact's display picture
+     * var user = blackberry.bbm.platform.users.contactsWithApp[0];
+     * var displayPicImg = document.getElementById("displayPicture");
```
ababut: We should put a comment here saying that we're assuming that the user has an  tag with id='displayPicture'

api/blackberry_bbm_platform_users_BBMPlatformUser.js
```diff
@@ -0,0 +1,96 @@
...
+     * displayPicImg.src = user.displayPicture;
```
ababut: Can both BB 5 and 6 take base-64 encoded data into the img tags? Some random blogger online was commenting that it only works in FF.

api/blackberry_bbm_platform_users_BBMPlatformUser.js
rburbidge, line 89: This is one of those concepts that is never easy to convey. I don't think that the docs for these can be improved any more. It already calls out the difference between them in bold.

api/blackberry_bbm_platform_users_BBMPlatformUser.js
```diff
@@ -0,0 +1,96 @@
...
+     * &lt;/script&gt;
+     * 
+     * @example
```
ababut: Duplicate example.

api/blackberry_bbm_platform.js
rburbidge: BBM6 is the version of BBM, which is independent of BBOS version. I will add something like "BBM6 is supported on OS5, OS6, and OS7"

api/blackberry_bbm_platform.js
rburbidge, line 56: Us generating it on their behalf could result in either inconsistent UUIDs or UUID collisions with other app developers. They need to know what their UUID is, it's super easy to generate one, and they only need to do it once.

Would it help if I put a link to an online UUID generator?

api/blackberry_bbm_platform.js
rburbidge, line 78: The UUID needs to be the same on different runs of the same application. If it changes then all of the sudden contacts who have my app installed will no longer appear to have it installed because it is a different UUID. We can't trust this to Math.random().

api/blackberry_bbm_platform.js
rburbidge: I could replace that with a generic "// Take action based on profile box item" comment.

api/blackberry_bbm_platform_users_BBMPlatformUser.js
rburbidge: They both work -- I've seen that the Contact.picture API works in both 5 and 6. However something is wrong with the way I base64 encode my images and it's currently not working in 5 with BBM API.

api/blackberry_bbm_platform_users_BBMPlatformUser.js
rburbidge: They aren't duplicates -- one shows display picture of current user, the other shows for a contact with the app. Which would you like me to keep?

api/blackberry_bbm_platform.js
ababut: Gotcha...in that case we should use branding-approved names. If BBM6 is it, ignore this comment, if not, pls find out what branding wants in there because we have to use that. For example, for instead of Playbook, they wanted our docs peppered with BlackBerry PlayBook :S

api/blackberry_bbm_platform.js
ababut, line 56: No, we shouldn't be endorsing 3rd party tools directly. 

I'm not sure, but can the Java UID generator be seeded with a String value to always produce the same UUID?

api/blackberry_bbm_platform.js
ababut, line 78: You're right, my bad!

api/blackberry_bbm_platform.js
ababut: Yes please...we want the sample code to not break anything when copy-pasted.

api/blackberry_bbm_platform_ui_menu.js
ababut: What happens if I have another app, completely different from yours, and I add menu items to a contact:
1: Solitaire
2: Golf
3: Poker

Do I overwrite your items?

api/blackberry_bbm_platform.js
rburbidge, line 56: I'd really like to leave this as is since this is how it works on the Java side. Assigning a UUID shouldn't be too much of a struggle for app developers.

api/blackberry_bbm_platform_ui_menu.js
rburbidge: Then those will show up under a different menu item... I should add a picture to the docs to show what the menu items look like.

api/blackberry_bbm_platform_users.js
ababut, line 75: What do they have to pass in for each of the categories?

api/blackberry_bbm_platform_users_BBMPlatformUser.js
ababut: Gotcha, I just saw that...I say we keep the self picture, the code example only references items on that docs page. The other one could be confusing because they might not know where contactsWithApp[0] came from.

api/blackberry_bbm_platform_users.js
rburbidge, line 75: I'm not quite sure what you mean.

api/blackberry_bbm_platform_users_BBMPlatformUser.js
ababut, line 76: Sounds good!

api/blackberry_bbm_platform_users_BBMPlatformUser.js
ababut, line 89: Hmm...how about something like:

handle - a unique, device-specific ID assigned to the user's account on each platform where it is used (BB, PB, etc)

ppid - a unique ID representing the user on all devices

rburbidge, line 89: That's actually very good. I'll play with that and see what I come up with.

api/blackberry_bbm_platform_users.js
ababut, line 75: You mentioned that they have to define users, type or both...so I'm guessing that the object literal you pass to pickUsers needs a type, users or both properties present.

When they are absent, what is their value assumed to be? We need to call out that at least one is required, and for the one not provided, what value it's assumed to have.

api/blackberry_bbm_platform_users_BBMPlatformUser.js
ababut: Heh, the BB 5 browser's always the joker on the deck. If your base-64 encoded image works fine in BB6 and 7 but not 5, I'd suspect the browser. 

If that's the case, I suggest getting rid of the example altogether. Either it works for everyone, or they all figure it out on their own and we show a different example :)

api/blackberry_bbm_platform_self.js
```diff
@@ -0,0 +1,92 @@
...
+         * the user to allow or deny the change.
+         * <p>The URI must be fully qualified. Non-local URIs must be whitelisted in the application's configuration file. Examples:
+         * <ul>
+         * <li>Locally from within the widget package (e.g. "local:///smiley.jpg")
+         * <li>From the filesystem of the device (e.g. "file:///SDCard/BlackBerry/pictures/smiley.jpg")
+         * </ul>
+         * </p>
+         * @param {String} displayPictureURI The fully qualified URI.
+         * @callback {Function} onComplete Invoked when the user has dismissed the dialog.
+         * @callback {Boolean} onComplete.result <code>true</code> if the user allowed the change;
```
ababut: Let's call the callback param "accepted" or "hasAccepted"...the example code would read more naturally.

api/blackberry_bbm_platform_self.js
```diff
@@ -0,0 +1,92 @@
...
+         *     }
+         * });
+         * &lt;/script&gt;
+         */
+        setDisplayPicture: function(displayPictureURI, onComplete) { },
+        
+        /**
+         * @description Sets the user's personal message. A dialog will be presented to the user to
+         * allow or deny the change.
+         * @param {String} personalMessage The personal message. The maximum length is 64 characters;
```
ababut, line 60: I suggest not mentioning the upper limit. It gives you freedom to change it in the future. This will encourage developers to add logic to check that the message isn't longer than 64 characters.

ababut, line 60: Also, if not too late, let's not encourage them to pass in null. Would empty string ("") also clear it?

api/blackberry_bbm_platform_self.js
ababut, line 75: Let's tag the parameter as optional then

api/blackberry_bbm_platform_self_profilebox.js
```diff
@@ -0,0 +1,192 @@
...
+ * <h2>Icons</h2>
+ * Before icons can be used in profile box items, a static set of icons must be registered using
+ * {@link blackberry.bbm.platform.self.profilebox.registerIcon}. An application should register its entire set of
+ * icons initially. This improves speed when the current user views this application in their
+ * contacts' profiles.</p> 
+ * 
+ * <h3>Icon IDs</h3>
+ * <ul>
+ * <li>Each icon must be given a unique non-negative ID.
+ * <li>Each icon ID must map to the same icon on all devices. Images generated at runtime such as
```
ababut: Is there a size constraint on images generated at runtime?

api/blackberry_bbm_platform_users_BBMPlatformUser.js
rburbidge: Yeah but as I said Contact.picture works in 5.0 and I'm convinced that I'll be able to get it working for my images too.

api/blackberry_bbm_platform_self.js
rburbidge: Good suggestion

api/blackberry_bbm_platform_self.js
rburbidge, line 60: If the text randomly gets truncated the devs will be angry because they weren't told why.

Why don't we want developers using null?

api/blackberry_bbm_platform_self.js
rburbidge, line 75: Will do

api/blackberry_bbm_platform_self_profilebox.js
rburbidge: No. Any images they provide will be scaled anyway.

api/blackberry_bbm_platform_users.js
rburbidge, line 75: The underlying java API has 3 overloaded methods: users, type, and users/type. So when one is missing it isn't really "defaulted" to anything, it just isn't used at all.

api/blackberry_bbm_platform_self_profilebox.js
ababut: Then can we use an image at runtime?

api/blackberry_bbm_platform_self_profilebox.js
ababut, line 65: Agreed...in the descriptions of the methods that throw the UserProfileBoxAccessException, we shouldn't mention that it's because the device is in MSM, rather that the Profile box is inaccessible. For the reasons why it might be inaccessible, they can refer to the description of this property and use it to test in code.

api/blackberry_bbm_platform_self_profilebox.js
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * The profile box items. <code>undefined</code> if the device is in mass storage mode.
+     * @type blackberry.bbm.platform.self.profilebox.ProfileBoxItem[]
+     * @readOnly
+     * @BB50+
+     */
+    items : null,
+    
+    /**
+     * Registers a new icon from a URI.
```
ababut: Is it possible to not impose the registration of icons at the beginning constraint? If it's for performance reasons, we can list it as a best practice, but registering an icon before adding it implies ordering of calls. 

What happens if they add an item and reference an ID that doesn't exist?

api/blackberry_bbm_platform_self_profilebox.js
ababut: Same comment here that I'm making everywhere that I see IDs: the extension should be managing them internally. I cannot see a reason to ask the JS developer to manage IDs.

api/blackberry_bbm_platform_self_profilebox.js
```diff
@@ -0,0 +1,192 @@
...
+     * @param {String} options.text The text of the item.
+     * @param {Number} [options.iconID] <i>Optional</i> The ID of a registered icon. If not provided
+     * then no icon will be used.
+     * @param {String} [options.cookie] <i>Optional</i> The customizable cookie.
+     * @returns {blackberry.bbm.platform.self.profilebox.ProfileBoxItem} the new item.
+     * @throws {UserProfileBoxAccessException} If the device is in mass storage mode.
+     * @example
+     * &lt;script type="text/javascript"&gt;
+     * // Add an item with icon, text, and a cookie
+     * var options = {text:"Planted an apple orchard on 10 acres!", iconId:ICON_INFO.apple.id, cookie:"10acres"};
```
ababut: ICON_INFO.apple.id is coming from the example for registerIcon. Since there several steps involved with updating the profileBox, we should consolidate all per-method examples into a single coherent example for the class. 

It will serve to bring together the process' various steps.

api/blackberry_bbm_platform_self_profilebox.js
rburbidge: Yes they can retrieve the icon using self.profilebox.getIcon()

api/blackberry_bbm_platform_users.js
```diff
@@ -0,0 +1,236 @@
...
+    /**
+     * @description Contacts who have the application installed.
+     * @type blackberry.bbm.platform.users.BBMPlatformUser[]
+     * @readOnly
+     * @BB50+
+     */
+    contactsWithApp : [],
+
+    /**
+     * @description Shows Contact Picker dialog allowing the user to select contacts.
```
ababut: This description needs to be updated - what are the picked users used for?

api/blackberry_bbm_platform_users.js
```diff
@@ -0,0 +1,236 @@
...
+     * }, "Did you see the game?");
+     * 
+     * &lt;/script&gt;
+     * @example
+     * &lt;script type="text/javascript"&gt;
+     * 
+     * // Start chat without Contact Picker dialog
+     * blackberry.bbm.platform.users.startBBMChat(function() {
+     *     // Continue with application...
+     * }, "I won by 500 points!", [player1, player2, player3]);
```
ababut: The players array will break this example.

api/blackberry_bbm_platform_users.js
```diff
@@ -0,0 +1,236 @@
...
+     * &lt;script type="text/javascript"&gt;
+     * 
+     * blackberry.bbm.platform.users.onUpdate = function(user, event) {
+     *     if (event == "displaypicture") {
+     *         // Get the img element for the user
+     *         var displayPicId;
+     *         var userHandle = user.handle;
+     *         if (userHandle == blackberry.platform.self.handle) {
+     *             displayPicId = "myPicture";
+     *         } else if (userHandle == p2Handle) {
```
ababut: pXHandle will break this example.

api/blackberry_bbm_platform_self_profilebox.js
rburbidge: The reason for registering before is so that I don't have to download icons if viewing the same app on another user's profile. It isn't absolutely necessary, but it's a better user experience.

If you reference the ID on a not-registered icon then it just shows up as no icon in the profile.

api/blackberry_bbm_platform_self_profilebox.js
rburbidge: Application developers can handle IDs.

api/blackberry_bbm_platform_self_profilebox.js
rburbidge: I can do that.

api/blackberry_bbm_platform_users.js
rburbidge: They can be used for anything. In my sample app I use pickUsers so that the user can select others to remove from a connection.

api/blackberry_bbm_platform_users.js
rburbidge: Then the only example I could really show is the top one.

Is it necessary that all examples must work if copy/pasted? Sometimes this will never be possible without straying from the point of the API. (case in point)

This may sound caustic, but I'm not sure if this audience of copy/paste developers is the kind we want to attract to our platform.

api/blackberry_bbm_platform_users.js
ababut: Gotcha, I understand now.

api/blackberry_bbm_platform_users.js
ababut: The issue is that an error in JavaScript immediately stops the script engine. To the user it looks like our code crashed their app. We had a number of situations when developers used our non-self-contained snippets, their app crashed and we received negative feedback. 

I totally understand the BBM API's scope, and I know a lot of pieces tie together, but the onus is still on us to provide non-crashing examples. This means combining several smaller examples into a big, class-level one if that makes sense. Or we can do what you did earlier and put comments replacing code. Whatever makes sense in a particular situation.

api/blackberry_bbm_platform_self.js
ababut, line 60: It leaves the door open to potential bugs coming from NullPointerExceptions. Empty Strings are safe.

It's up to you for the message, but telling people the maximum length will get them coding against it.

api/blackberry_bbm_platform_self_profilebox.js
ababut: Java application developers handle IDs routinely :) JavaScript developers do not.

Typically, in JS you pass around an object that opaquely hides the implementation and has a few basic use-case oriented methods on it.

api/blackberry_bbm_platform_ui_menu.js
ababut: Yes please, or just update the description.

api/blackberry_bbm_platform_users_ContactInvitation.js
jeffheifetz: Well why are they split then. If you'd group the functionality under the TOC, it feels like it should be grouped under one namespace then.

rburbidge: I was incorrect about that. They aren't split. The two methods are users.inviteToBBM and users.inviteToBBMFromConnections.

I tried to put a TOC entry on the two methods but it doesn't show up. Is there a way to put TOC entries for methods?

rburbidge:
About the tables: I can always add the class="confluenceTh" and class="confluenceTd" attributes to each cell in the tables. Although it still leaves to docs open for inconsistency if you ever choose to change the style of the docs.

Alternatively I could use lists with the item being described in bold.

api/blackberry_bbm_platform_users.js
rburbidge: I will remove the second example.

Issue #192, opened 2011-06-24
genDoc.bat does not delete template folder prior to xcopy


cdelcol:
If you move any of the template files, the original name/location persists because the destination folder is not removed prior to the xcopy in the script

PR #193, opened 2011-06-24
Fix for #192: gendocs now deletes the destination template directory before xcopy

cdelcol:


cdelcol:
Sent pre-maturely. Needs to be flattened and re-submitted.

PR #195, opened 2011-06-24
Fix for #192: gendocs now deletes the destination template directory before xcopy

cdelcol:
Fix allows for correctly moving/renaming/removal of files in the docs templates directory.


ababut:
Looks good, please pull this in.

Issue #198, opened 2011-06-24
File api : remove PlayBook support and sample code


cdelcol:
Methods:
blackberry.io.file.saveFile 
blackberry.io.file.readFile 

Sample code:

<script type="text/javascript">
  //This example is for PlayBook it assumes that a test.jpg file exists in the photos directory
  //Substitute the file name as necessary

  var dirs = blackberry.io.dir.appDirs;
  // readFile
  var globalBlob;
  function onFileOpen(fullPath, blobData){
      globalBlob = blobData;
  }

  try{
      blackberry.io.file.readFile(dirs.shared.photos.path + "/test.jpg", onFileOpen, true);
  }catch(e){
      alert(e);
  }

  //saveFile
  var newFilename = "/testCopy.jpg";
  var newFilePath = dirs.shared.photos.path + newFilename;
  blackberry.io.file.saveFile(newFilePath, globalBlob);
</script>

cdelcol:
Without string to blob/blob to string, these are useless.

cdelcol:
Resolved in master branch.

PR #200, opened 2011-06-27
Fix for issue #198 read/save file on Playbook removed

cdelcol:
A fix that should make it into master (we may re-spin to pick up a few fixes in the delta release) and be merged up into the epsilon branch as well.


ababut:
Looks good, please merge.

Issue #194, opened 2011-06-24
README.md for the Build folder


tneil:
We should place a README.md in the build folder so that people can understand how to use the genDoc.bat

cdelcol:
Fixed with issue #208.

PR #208, opened 2011-07-02
bash Script for generating docs

nukulb:
bash script for generating the docs. It copies the template and the plugin and then calls the jsrun.jar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,11 @@ This project is Open Sourced under the Apache 2.0 license
 
 [Read more](https://github.com/blackberry/WebWorks/wiki) about the BlackBerry WebWorks open source project
 
+## Automated Build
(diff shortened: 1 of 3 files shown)


tneil:
Is this the same as the genDoc.bat for Windows but now also for Mac?  If they are intended to be the same, can we attempt to make the input parameters (if any) the same between the bat and the bash script?

tneil:
Could we also add the following:
- README.md file for the /build directory that talks about how to use both the genDoc.bat and genDoc bash files
- updated the main root repo README.md that outlines that there are also mac/windows build scripts that can be found in the /build directory of the repo

tneil:
I have an associated feature Issue of #194

nukulb:
Its modeled after the genDoc.bat file so the input parameters are the same.

There is a usage included in bash scripts which talks about how to use it.

I will add both the README.md and make the note on the main page.

Do you want me to add this as a part of the same pull request?

tneil:
Sure.. just make it part of the same pull request

nukulb:
Took some editing in github to get it to format correctly but I think what I have now works.

nukulb:
I need to find a way to have them as relative paths.
I will update.

cdelcol:
Manually merged pull request.

PR #218, opened 2011-07-19
Jsdoc source for BlackBerry Advertising

ssavla:
Updated source for the review

### DIFF
--- a/api/blackberry_advertising.js
+++ b/api/blackberry_advertising.js
@@ -16,7 +16,7 @@
 
 /**
  * @beta <a href="https://adservices.blackberry.com/register">Sign up</a> for the BlackBerry&reg; Advertising Service <br>
- * <a href="#placeholder"> Development Guide - Advertising Service for BlackBerry WebWorks </a>
+ * <a href="http://docs.blackberry.com/en/developers/deliverables/34109/index.jsp?name=Development+Guide+-+Advertising+Service+for+BlackBerry+WebWorksAdvertising+Service1.0+Beta&language=English&userType=21&category=BlackBerry+WebWorks+for+Smartphones&subCategory="> Development Guide - Advertising Service for BlackBerry WebWorks </a>
  * @toc {Advertising} Banner
  * @featureID blackberry.advertising
  * @class The Banner object is an advertising banner that will be created and automatically added into a specified div in your html.
--- a/api/blackberry_advertising.js
+++ b/api/blackberry_advertising.js
@@ -1,0 +1,125 @@
+/*
+* Copyright 2010-2011 Research In Motion Limited.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+/**
+ * @toc {Advertising} Banner
+ * @featureID blackberry.advertising
+ * @class The Banner object is an advertising banner object that will automatically get inserted into a div in your html.
+ * @constructor
+ * @description Creates a new Banner Object with the given properties and automatically inserts it into the div with the specified id.
+ * @param {Number} appID The Application ID given to the developer by the BBAS on registration.
+ * @param {String} divID The id attribute of the div tag to which the banner will be inserted.
+ * @param {JSON} [properties] A set of options to further customize the advertising banner.
+ * @param {Boolean} [properties.bannerTransitions] Sets the {@link blackberry.advertising#bannerTransitions} property.
+ * @param {String} [properties.focusBorderColor] Sets the {@link blackberry.advertising#focusBorderColor} property.
+ * @param {String} [properties.focusBorderWidth] Sets the {@link blackberry.advertising#focusBorderWidth} property.
+ * @param {JSON} [properties.metadata] Sets the {@link blackberry.advertising#metadata} property.
+ * @param {String} [properties.mmaSize] Sets the {@link blackberry.advertising#mmaSize} property.
+ * @param {String} [properties.placeholder] Sets the {@link blackberry.advertising#placeholder} property.
+ * @param {Number} [properties.refreshRate] Sets the {@link blackberry.advertising#refreshRate} property.
+ * @BB60+
+ * @example
+ * &lt;center&gt; &lt;div id='bannerDiv'&gt;     &lt;/div&gt; &lt;/center&gt;
+ * &lt;script type="text/javascript"&gt;
+ * 
+ *  //Adding Banner with default Banner properties
+ *  var banner1 = banner(16126, "bannerDiv");   
+ * 
+ *  //Adding Banner with custom Banner properties
+ *  var banner1 = new blackberry.advertising.Banner(16126, "bannerDiv", {"bannerTransitions" : true,
+ *                                                                 "mmaSize" : "extralarge",
+ *                                                                 "focusBorderColor" : "blue",
+ *                                                                 "focusBorderWidth" : "2px",
+ *                                                                 "refreshRate" : 6000,
+ *                                                                 "placeholder" : "http://test.com/placeholder.jpg"},
+ *                                                                 "metadata": {"age" : "12",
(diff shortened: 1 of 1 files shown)


ssavla:
use the last commit for the review

api/blackberry_advertising.js
```diff
@@ -0,0 +1,143 @@
...
+ * &lt;xml version="1.0" encoding="UTF-8" ?&gt;
+ * &lt;widget xmlns="http://www.w3.org/ns/widgets" xmlns:rim="http://www.blackberry.com/ns/widgets" version="2.0" rim:header="RIM-Widget: rim/widget"&gt;
+ * &lt;name&gt;App Name&lt;/name&gt;
+ * &lt;access subdomains="true" uri="http://www.blackberry.com/"&gt;
+ * &lt;feature id="blackberry.advertising" required="true" version="1.0.0" /&gt;
+ * &lt;/access&gt;
+ * &lt;access subdomains="false" uri="*" /&gt;
+ * &lt;content src="index.html" /&gt;
+ * &lt;rim:loadingScreen foregroundImage="loader.gif" onRemotePageLoad="true" onLocalPageLoad="true" onFirstLaunch="true" /&gt;
+ * &lt;/widget&gt;
```
jeffheifetz: This information should be moved into the class description or maybe an @notice tag and placed between < code>  tags.
Also a cleaner config.xml without all the extra information would be more useful.

api/blackberry_advertising.js
```diff
@@ -0,0 +1,143 @@
...
+ *                                                                 "focusBorderColor" : "blue",
+ *                                                                 "focusBorderWidth" : "2px",
+ *                                                                 "refreshRate" : 6000,
+ *                                                                 "placeholder" : "http://test.com/placeholder.jpg"},
+ *                                                                 "metadata": {"age" : "12",
+ *                                                                              "gender" : "male",
+ *                                                                              "city" : "SanFrancisco"}
+ *                                                                };
+ *  
+ * &lt;/script&gt;
```
jeffheifetz, line 79: Step 2/3 can be combined into one code example with both HTML and JS. This example should work out of the box if the user simply copies and pastes it into an html file.

api/blackberry_advertising.js
```diff
@@ -0,0 +1,143 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+/**
+ * @toc {Advertising} Banner
+ * @featureID blackberry.advertising
```
jeffheifetz, line 21: You need to add an @beta tag.

api/blackberry_advertising.js
ssavla: Thanks Jeff.. I will remove name, content and loading tags from config.xml
and use <code> tags ..
I have one question regarding @notice and class description, currently this sample code shows up at the end of API description. So if I use 'class description' or 'notice' will it change the position of this section ?

api/blackberry_advertising.js
ssavla, line 79: Ok.

api/blackberry_advertising.js
ssavla, line 21: sure

api/blackberry_advertising.js
jeffheifetz: Yes and this was why I wanted you to do so. The class description (@class) will show at the top of the page
The @notice tag would also be at the top but in bright yellow

ssavla: Jeff, do you want me to move just the text "Set access to "blackberry.advertising" and "www.blackberry.com" in the app config.xml"
or even the XML code in the notice tag? 

I tried to put code within notice .. but it does not look good. Basically code is being displayed as a continuous paragraph.
Rather shall I move the XML code into example section ? and make notice as,
 "Set access to blackberry.advertising and www.blackberry.com in the app config.xml, see the example"

ssavla: < br > tag helps .. i have emailed you the screenshot of the updated doc ... so that we are on same page

api/blackberry_advertising.js
```diff
@@ -0,0 +1,142 @@
...
+ * Minimum requirement is BlackBerry&reg; WebWorks SDK 2.1.<br>
+ * Set access to "blackberry.advertising" and "www.blackberry.com" in the app config.xml
+ * <br> <br>
+ * <code>
+ * &lt;xml version="1.0" encoding="UTF-8" ?&gt; <br>
+ * &lt;widget xmlns="http://www.w3.org/ns/widgets" xmlns:rim="http://www.blackberry.com/ns/widgets" version="2.1" rim:header="RIM-Widget: rim/widget"&gt; <br>
+ *   &nbsp;&nbsp;&nbsp;&lt;access subdomains="true" uri="http://www.blackberry.com/"&gt; <br>
+ *     &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&lt;feature id="blackberry.advertising" required="true" version="1.0.0" /&gt; <br>
+ *   &nbsp;&nbsp;&nbsp;&lt;/access&gt; <br>
+ * &lt;/widget&gt; <br>
```
jeffheifetz, line 34: You're missing the <code> < accesss subdomains="true" uri="*" /> </code>

api/blackberry_advertising.js
```diff
@@ -0,0 +1,142 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+/**
+ * @beta
```
jeffheifetz: I think you should put a link to the relevant Ad SDK resources in the Beta tag

api/blackberry_advertising.js
```diff
@@ -0,0 +1,142 @@
...
+ *                                                                 "focusBorderColor" : "blue",
+ *                                                                 "focusBorderWidth" : "2px",
+ *                                                                 "refreshRate" : 6000,
+ *                                                                 "placeholder" : "http://test.com/placeholder.jpg"},
+ *                                                                 "metadata": {"age" : "12",
+ *                                                                              "gender" : "male",
+ *                                                                              "city" : "SanFrancisco"}
+ *                                                                };
+ * &lt;/script&gt;
+ * </code>
```
jeffheifetz: I don't believe the heading or the code tags are necessary, these should be implied by it being an example.

api/blackberry_advertising.js
```diff
@@ -0,0 +1,142 @@
...
+ * @example
+ * Include Advertising SDK JavaScript and API code into all the HTML pages as shown below:
+ * <code>
+ * &lt;script type="text/javascript" src="http://www.blackberry.com/app_includes/asdk/adBanner.js"&gt;&lt;/script&gt;
+ *  
+ * &lt;center&gt; &lt;div id='bannerDiv'&gt;     &lt;/div&gt; &lt;/center&gt;
+ *
+ * &lt;script type="text/javascript"&gt;
+ *  //Adding Banner with default Banner properties
+ *  var banner1 = new blackberry.advertising.Banner(16126, "bannerDiv");
```
jeffheifetz: I don't think this line is necessary in the example, but if you want to show I'd prefer a second @example tag so both are usable on their own.

api/blackberry_advertising.js
jeffheifetz: Nice job on formatting the notice, seems fine to me.

api/blackberry_advertising.js
ssavla, line 34: done.

api/blackberry_advertising.js
ssavla: will remove tags

api/blackberry_advertising.js
ssavla: Good catch! What about having one more div for the second 'banner' call ?
That way example will be valid and usable

ssavla: Hmm on second thought, I think separate example will be less confusing and cleaner :)

jeffheifetz: You're call really as both are fine, but I agree that separate examples makes more sense to me.

jeffheifetz: The two examples are already in the example section, I do not believe they need EXAMPLE headers or any other text but the example itself.

cdelcol:
Please see blackberry-webworks/WebWorks-API-Docs#29

PR #222, opened 2011-07-22
Added BB7.0 to the drop down filter and included APIs supported on BB6.0

jachoi:


jeffheifetz:
Do you think we could update the code to generate all of the headers and the filters so that we only ever have to change this in one place in the future?

jachoi:
Are there other headers and/or filters that I will be needing to put in? I was only told to put in BB7.0 filter in the drop down menu.

cpinzon:
Jeff, what would you suggest in terms of an update of code to avoid changing this all over the place once we need a new header and filter?

jeffheifetz:
Well we've already mostly centralized the logic to that BBTag class so I would think make some more methods there. At the very least only the BBTag class will need to change.

jeffheifetz:
Also you're missing the changes needed in the featureList.tmpl. They should be identical to those in the Toc

cpinzon:
Yup, sorry about that James, didn't catch that. Jeff, makes sense. I can bring this to Peter's attention.

jachoi:
Sure if you can that would be nice. Thanks.

cpinzon:
James, can you add the code in the featureList.tmpl file and push it so it can be added to the code review?

jachoi:
Done adding the code in featureList.tmpl file.

cpinzon:
This looks good. I generated the docs and no apparent problems.

cdelcol:
closed via commit #0a86198

PR #225, opened 2011-07-27
(PIM) Category API Signature Changes?

cdelcol:
Reference: http://www.blackberry.com/developers/docs/webworks/api/blackberry.pim.category.html

In the docs, the Category API reference has changed from:

```
blackberry.pim.addCategory
```

to:

```
blackberry.pim.category.addCategory
```

Note: the code sample still references the first api signature.

Were the API definitions mistakenly changed in the docs, or has this change actually been implemented in WebWorks itself?

Cheers,

### DIFF
--- a/api/blackberry_pim_category.js
+++ b/api/blackberry_pim_category.js
@@ -23,12 +23,6 @@
 * @BB50+
 * @namespace The Category object defines APIs necessary to access and manipulate categories in the PIM  database.
 * @featureID blackberry.pim.category
-* @example
-* &lt;script type=&quot;text&sol;javascript&quot;&gt;
-*   function createNewCategory(newCategoryName) {
-*     blackberry.pim.addCategory(newCategoryName);
-*   }
-* &lt;&sol;script&gt;
 */
 blackberry.pim.category = { };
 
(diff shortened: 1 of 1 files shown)


cdelcol:
fixed the namespace in the sample for correctness
Moved the sample under the proper method for clarity

brentlintner:
Hmm, I'm a bit confused with the recent pull request. Does this mean that the API structure did change for category at some point?

My Java is very rusty, but going off master on blackberry/WebWorks, it seems the code still only exports functions that get appended to the blackberry.pim object?

https://github.com/blackberry/WebWorks/tree/restructure/api/pim/src/main/java/blackberry/pim/category

cdelcol:
From brentlintner:

Hey Chris,

Thanks. The main confusion for me was I know the original docs (before jsdocs) actually referenced pim.addCategory and not pim.category.addCategory, since "pim.category" is the feature URI in WebWorks, but the actual API methods were in the blackberry.pim object. 

The pull request (unless I was mistaken) seemed to fix things in favour of the "blackberry.pim.category.addCategory" signature instead of the other way around (like it used to be).

cdelcol:
Verified that code implementation matches fixed docs. We should be good.

PR #228, opened 2011-08-02
Tablet fixing details for get existing purchases on error callback

tohman:


cdelcol:
Manually merged into master.

PR #249, opened 2011-08-22
Tablet blob utils methods supported in play book

tohman:


tohman:
Mistake, please ignore.

PR #271, opened 2011-10-28
BBM fixes

rburbidge:
Fixed some broken links, incorrect parameter types, and added better link to getting started guide.


rburbidge:
Bah I think I put this in the wrong place.

PR #316, opened 2012-05-29
Next bb10 dialog update

jengee:
fixes issue #122

https://github.com/blackberry-webworks/WebWorks-API-Docs/issues/122


ericpearson:
I think maybe the documentation should say that the dialog appears on top of all other applications as well. At least that was my understanding of the flag.

jengee:
incorrect repo request. New pull request created here: https://github.com/blackberry-webworks/WebWorks-API-Docs/pull/123

jengee:
submission is screwed up. Will resubmit pull request

PR #353, opened 2012-09-12
The bb.io read from HTML5 FileySystem example does not work.

brentlintner:
Ref: https://developer.blackberry.com/html5/apis/blackberry.io.html
- The file may not exist, so {creat: true} should be in there.
- `documents` folder should be capitalized.

Note: Let me know if I am mistaken about `Documents`, but I understand it _is_ capitalized, and the file system is case sensitive.


jeffheifetz:
Seems right to me, thanks. r+

PR #368, opened 2012-10-16
blackberry.pim.message API

EricLeiLi:
@kwallis @nukulb @rwmtse 
all the interfaces for blackberry.pim.message are defined, please take a look.


nukulb:
@EricLeiLi  it should be going into next-BB10, wrong base branch. Please fix.

nukulb:
Its hard to read the docs since I can't see them built, for some reason they are appearing in browser.

Here are the used cases I couldn't find-
1. How will the user create an email listener? So if I want to listen to incoming email, how do I do it?
2. How will the user listen to changes to a message? So if I create a draft and put it in the drafts folder and now I want to listen to changes to this message. Can I do it?

PR #397, opened 2012-12-12
Adding platform overriding + clearing abilities to CCM API.

ejzn:


ejzn:
@ishneur, @nukulb retargeted.

PR #398, opened 2012-12-12
Updates email composer and event picker docs to mention that local:/// p...

haahmad:
...aths are not supported.

### DIFF
--- a/api/blackberry_invoke_card.js
+++ b/api/blackberry_invoke_card.js
@@ -464,7 +464,7 @@ blackberry.invoke.card = {
         invokeTargetPicker : function(options, title, onSuccess, onError) {},
 
         /**
-         * @description Invokes calendar event picker card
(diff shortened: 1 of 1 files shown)


api/blackberry_invoke_card.js
```diff
@@ -464,7 +464,7 @@ blackberry.invoke.card = {
         invokeTargetPicker : function(options, title, onSuccess, onError) {},
 
         /**
-         * @description Invokes calendar event picker card
+         * @description Invokes calendar event picker card. <b>Note</b>: "local:///" filepaths are not supported for this card.
```
jeffheifetz, line 467: I believe there is an `@apinotice` tag, but I could be wrong.

jeffheifetz:
Also, please re-target to master since this is for in-market release.

PR #403, opened 2012-12-13
Added context menu disablePlatformItem/enablePlatformItem/listDisabledPl...

rcruz:
...atformItems

### DIFF
--- a/api/blackberry_ui_contextmenu.js
+++ b/api/blackberry_ui_contextmenu.js
@@ -80,7 +80,7 @@ blackberry.ui.contextmenu = {
  *
  */
  addItem : function (){},
- 
+
  /**
  * @description Allows the developer to remove previously added custom items from the context menu.
  * @param {String[]} contexts An array of constants defining which contexts this new item should appear.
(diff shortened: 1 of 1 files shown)


ejzn:
I think this somehow needs to be rebased. @nukulb Did we sort out what happened with that branch? This one has my change too...

api/blackberry_ui_contextmenu.js
```diff
@@ -134,6 +134,127 @@ removeItem: function (){},
 defineCustomContext: function (){},
 
 /**
+ * @description Allows the developer to override platform menu items in the context menu.
+ * @param {Object} action An object that should be used to override the platform menu item using the actionId.
+ * @param {String} action.actionId A property that uniquely defines the action to be added to the context menu.
+ * @param {String} action.label A string that will be displayed to the user describing the custom context to be performed. ie: Edit
+ * @param {String} action.icon A path to the an image to display in the context menu
+ * @callback {Function} callback A function which runs when the menu item action is executed
```
jeffheifetz, line 142: Perhaps a name like onComplete would be more descriptive than callback

api/blackberry_ui_contextmenu.js
```diff
@@ -134,6 +134,127 @@ removeItem: function (){},
 defineCustomContext: function (){},
 
 /**
+ * @description Allows the developer to override platform menu items in the context menu.
+ * @param {Object} action An object that should be used to override the platform menu item using the actionId.
+ * @param {String} action.actionId A property that uniquely defines the action to be added to the context menu.
+ * @param {String} action.label A string that will be displayed to the user describing the custom context to be performed. ie: Edit
+ * @param {String} action.icon A path to the an image to display in the context menu
```
jeffheifetz, line 141: Are any of the params optional? If so they should be surrounded by []'s

api/blackberry_ui_contextmenu.js
```diff
@@ -134,6 +134,127 @@ removeItem: function (){},
...
+ *     blackberry.ui.contextmenu.enablePlatformItem(context, actionId);
+ * }
+ *
+ * &lt;/script&gt;
+ */
+ enablePlatformItem : function (){},
+
+/**
+ * @description Lists the disabled platform menu items in the context menu.
+ * @return {Object[]} 2D array of disabled items.  First index is context, second index is actionId
```
jeffheifetz, line 234: Shouldn't this be Object[][]. Also, do we have a type for ContextItems that we can put here instead of Object?

jeffheifetz:
This is waiting on 10.2, correct?

ejzn:
Well not exactly. It could be folded, just would have some strangeish behaviour in WP.

PR #405, opened 2012-12-14
Adding the sourceId property to Custom Context to facilitate getting the...

ejzn:
... source element in the callback


api/blackberry_ui_contextmenu.js
```diff
@@ -97,7 +97,8 @@ blackberry.ui.contextmenu = {
  * @param {String} action.actionId A property that uniquely defines the action to be added to the context menu.
  * @param {String} action.label A string that will be displayed to the user describing the custom context to be performed. ie: Edit
  * @param {String} action.icon A path to the an image to display in the context menu
- * @callback {Function} callback A function which runs when the menu item action is executed
+ * @callback {Function(sourceId)} callback A function which runs when the menu item action is executed
```
jeffheifetz: I don't believe this type is consistent with how we document callbacks. Do you prefer the layout this way?

ejzn:
I'm not sure what the alternative is in this case. JSDocs doesn't provide a clean mechanism of documenting the input paramters to a callback. This way at least it does document it properly here is what it looks like:
sourc_id

api/blackberry_ui_contextmenu.js
jeffheifetz: I meant that because you made the type `Function(sourceId)` in your output you know see `Function(sourceId)(sourceId: String)`. I would prefer it to just have type `Function` and that way it wouldn't have two sets of brackets.

ejzn: Sure that is an easy change

Issue #406, opened 2012-12-15
Missing permission in Geolocation HTML5 API


oparoz:
For the GPS to be accessible from WebWorks apps, an extra permission is required:
access_location_services

This is missing from the docs.

bryanhiggins:
Thanks for pointing this out. The permit has migrated from read_geolocation to access_location_services.

PR #407, opened 2012-12-16
Update permit for HTML5 geolocation

bryanhiggins:
Issue #406


nukulb:
Not true for playbook

bryanhiggins:
Good point! I amended the commit. Let me know if it is ok - I didn't see other examples of multiple permit attributes.

jeffheifetz:
Thats because we have never had this come up before. Currently there is a way to specify that a featureId is valid for only a given set of OSes (using the  type {} option). 

Would we like to add this ability for permissions?

bryanhiggins:
I suggest we merge this into master ASAP.

It seems to be causing confusion in the community.

https://github.com/blackberry/BB10-WebWorks-Framework/issues/487

jeffheifetz:
Issue tracked in #419, r+

jeffheifetz:
Woops, needs to be retargetted to master

PR #411, opened 2012-12-18
Updates cards docs to indicate that local paths are not supported

haahmad:


api/blackberry_invoke_card.js
```diff
@@ -464,7 +464,7 @@ blackberry.invoke.card = {
         invokeTargetPicker : function(options, title, onSuccess, onError) {},
 
         /**
-         * @description Invokes calendar event picker card
+         * @description Invokes calendar event picker card. <b>Note</b>: "local:///" filepaths are not supported for this card.
```
jeffheifetz: Check out `@apiNotice`, there's an example in blackberry.app.events

haahmad: This is not really a change in the API. Local paths have never worked for these cards. Should I still use an `@apiNotice` tag?

jeffheifetz:
Also, please make sure this branch is based on blackberry-webworks:master

PR #417, opened 2012-12-18
Correct BB5+ event example

mcfedr:
I am not 100% sure this is right, but in my testing today it seems that on the pre BB10 webworks, event listeners have to be set by assigning a function to the relevant property. This seems to have been confused throughout the documentation, this just happens to be the bit that I was trying to get working today.

### DIFF
--- a/api/blackberry_bbm_platform.js
+++ b/api/blackberry_bbm_platform.js
@@ -97,7 +97,7 @@ blackberry.bbm.platform = {
      * &lt;script type="text/javascript"&gt;
      * 
      * // Create callback invoked when access changes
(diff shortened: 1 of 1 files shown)


jeffheifetz:
Looks like a great fix to me, but this branch can't be auto-merged. Can you rebase off of blackberry:master?

jeffheifetz:
Ignore that last rebase part, I will merge

jeffheifetz:
Now in the repo- https://github.com/blackberry/WebWorks-API-Docs/commit/e01d9f76e81fcbf051ada82e6d5047dce5f220b8

PR #424, opened 2012-12-21
The PPG URL for push needs to be http not https.

mdandrea:


nukulb:
merged manually, it will go out with the next release (timelines TBD, in heavy discussions right now)

PR #425, opened 2012-12-24
Update sample for cardStartPeek method

tohman:
In sample for cardStartPeek call this method after finger is placed on a DOM element.


pelegri:
Looks good to me, but I'm sure more qualified folks will show up next week and will review it :)

PR #430, opened 2013-01-09
Edit samples and signature description for 'directory' option

tohman:
Passing multiple directories in options for FilePicker is not supported now. When passing the list of directories only the first item in array will have an effect the other items would be ignored.


tohman:
@EricLeiLi review?

nukulb:
merged manually

PR #442, opened 2013-01-18
Updating localstorage to mention 25MB limit on BB10

jeffheifetz:


kwallis:
r+

jeffheifetz:
This is 25mb per domain, correct?

kwallis:
Good question, you may need to confirm.

---

This transmission (including any attachments) may contain confidential information, privileged material (including material protected by the solicitor-client or other applicable privileges), or constitute non-public information. Any use of this information by anyone other than the intended recipient is prohibited. If you have received this transmission in error, please immediately reply to the sender and delete this information from your system. Use, dissemination, distribution, or reproduction of this transmission by unintended recipients is not authorized and may be unlawful.

jeffheifetz:
Confirmed with otto cheung, updating commit

jeffheifetz:
Merged Manually

PR #447, opened 2013-01-24
Added blackberry.payment BB10 api

rwmtse:


api/blackberry_payment.js
```diff
@@ -94,42 +188,189 @@ blackberry.payment = {
...
+    *      alert("Error occurred: " + error.errorText + ", " + error.errorID);
+    *  }
+    * &lt;/script&gt;
+    */
+    checkExisting : function(skuOrId, onSuccess, onFailure) {
+    },
+    /**
+    * @function
+    * @description Initiates a request to cancel a subscription for a digital good.  If successful the subscription becomes
+    * canceled and will not be renewed at the end of the current subscription period.  The subscription is remains valid until
```
scdowney: Looks like a typo after changes. Should be something like: "The subscription will remain valid until the end of the current subscription period"

rwmtse: Thanks for spotting that.

api/blackberry_payment.js
```diff
@@ -32,25 +32,61 @@ blackberry.payment = {
 
     /**
      * @function
+     * @name blackberry.payment.getExistingPurchases^2
+     * @description Retrieves the previous successful purchases made by the user from within the calling application.
+     * @param {Boolean} refresh True if the BlackBerry should be allowed to refresh the list of purchases from the Payment Service server. False if the current list of cached purchases should be returned immediately.
```
scdowney: We want to change this message for clarity moving forward. It is preferred if we say something like "True in order to force a server refresh, false if the cached list is acceptable". The difference is minor, but it clarifies a lot in terms of what bps and paymentsystem are actually doing in the back end compared to the impression the documentation makes. Basically, if serverRefresh is true, it is absolutely forced, false will return a cached list if available (but it MAY still have to go to the server in some cases).

rwmtse: Can you give the exact description and I'd replace it. What I have on there is from the original BB WW api.

scdowney: Sure thing. Lets say:

True if the BlackBerry should retrieve the list of purchases from the Payment Service server. False if the BlackBerry should attempt to use the local purchase history, and only go to the Payment Service server if necessary.

rwmtse: Thanks. I've changed to your description.

api/blackberry_payment.js
```diff
@@ -32,25 +32,61 @@ blackberry.payment = {
 
     /**
      * @function
+     * @name blackberry.payment.getExistingPurchases^2
+     * @description Retrieves the previous successful purchases made by the user from within the calling application.
+     * @param {Boolean} refresh True if the BlackBerry should retrieve the list of purchases from the Payment Service server. False if the BlackBerry should attempt to use the local purchase history, and only go to the Payment Service server if necessary.
+     * @callback {function} onSuccess Function to be invoked on successful call.
+     * @callback {Array} onSuccess.data An array of {@link Purchase} items
```
jeffheifetz: Shouldn't the type be Purchase[] ?

rwmtse: Good point, let me fix that.

jeffheifetz:
Have the docs not updated correctly? developmentMode seems to be missing BB10 support, but I see it in the file.

scdowney:
Unfortunately, our SVV has uncovered another issue that must be addressed before we are comfortable with this going live. We are going to be investigating it tomorrow morning.

rwmtse:
@jeffheifetz Were you looking at the github copy? Try this [link](http://wwci:8080/view/WebWorks-API-Docs/view/Jobs/job/WebWorks-API-Docs-next-BB10-payments-rt/API_Reference/blackberry.payment.html#.developmentMode) instead

api/blackberry_payment.js
```diff
@@ -94,42 +188,189 @@ blackberry.payment = {
...
+    * @callback {function} onSuccess Function to be called on success.
+    * @callback {Boolean} onSuccess.data.subscriptionExists Whether the subscription digital good is currently active.
+    * @callback {function} onFailure Function to be called when an error occurs.
+    * @callback {PaymentError} onFailure.error A {@link PaymentError} object containing details about the error.
+    * @BB10X
+    * @example
+    * &lt;script type="text/javascript"&gt;
+    *   function checkSubscription() {
+    *       try {
+    *           blackberry.payment.checkExisting("12345", onSuccess, onFailure);
```
scdowney: We need to make a change to this API call, because of a limitation on the paymentsystem service that was just identified, and will not be fixed for release. Ash is implementing the fix in the payments extension, but the checkExisting call needs to take two separate parameters, id AND sku, and the documentation needs to be updated to reflect the following:

Only one of ID or SKU needs to be provided, and if both are provided, the ID takes precedence and SKU will be ignored.

scdowney:
Ash has pushed the fixed extension implementation to next-bb10-payments branch we had been working out of, but the documentation will need to be updated, as well as the unit tests, I believe.

api/blackberry_payment.js
```diff
@@ -94,42 +188,189 @@ blackberry.payment = {
...
+    *  function onSuccess(data) {
+    *      alert("Cancellation " + (data.subscriptionCancelled ? "" : "NOT ") + "successful");
+    *  }
+    *
+    *  function onFailure(error) {
+    *      alert("Error occurred: " + error.errorText + ", " + error.errorID);
+    *  }
+    * &lt;/script&gt;
+    */
+    cancelSubscription : function(transactionID, onSuccess, onFailure) {
```
scdowney: all references to "transactionID" actually need to be changed to "purchaseID" in cancelSubscription.

api/blackberry_payment.js
```diff
@@ -149,6 +390,7 @@ Purchase = { };
  * @description The transaction ID for this purchase.
```
scdowney, line 407: We also need to add purchaseID to the Purchase object documentation. It is already being populated on the output, just not documented, it appears.

rwmtse:
@scdowney how about getPrice? I see that it takes skuOrId ...?

scdowney:
That one is fine, actually. The implementation on the back end there doesn't have the same limitation                                                                                                                                      Sent from my BlackBerry 10 smartphone.                                                                                                                                                                                                                           From: Rosa TseSent: Wednesday, January 30, 2013 3:36 PMTo: blackberry/WebWorks-API-DocsReply To: blackberry/WebWorks-API-DocsCc: Shawn DowneySubject: Re: [WebWorks-API-Docs] Added blackberry.payment BB10 api (#447)@scdowney how about getPrice? I see that it takes skuOrId ...?

```
          —
          Reply to this email directly or view it on GitHub.
```

rwmtse:
@scdowney: I see. But I think we should unify the signature... coz outside developers would only look at the signature

scdowney:
@rwmtse Yeah, I agree. nice to keep it consistent between the two. we'll just have to do a bit more testing once the change is made, just to be absolutely sure getprice continues to work properly (though theres no reason why it shouldnt)

scdowney:
will these have build anywhere yet? I wouldnt mind seeing them in the rendered form... its easy to miss stuff when just looking at code

scdowney:
Just as a side note, our SVV are currently updating their app and testing the new changes, but Ash and I have also gone through a few tests on the new changes, and everything _seems_ to be working OK. I can update more definitively once I hear from SVV, but its looking good

rwmtse:
See the rendered docs [here](http://wwci:8080/view/WebWorks-API-Docs/view/Jobs/job/WebWorks-API-Docs-next-BB10-payments-rt/API_Reference/blackberry.payment.html)

scdowney:
Awesome, thanks @rwmtse. Docs all look good to me, now. And SVV are just running tests now, I am told.

Just curious, what is the timeline for this once both pull requests are merged in, in terms of public release?

jeffheifetz:
Merged Manually

PR #462, opened 2013-05-07
Documentation updates for PSDK error code improvements

scdowney:
Adding errorInfo attribute to support underlying error code improvements outlined here:
http://wikis.rim.net/display/bbaw/PSDK+Error+Code+Improvements

@bryanhiggins identified that this may not be the correct place to put docs changes anymore, but ensured they would get to the correct place. 

NOTE: this depends on 10.2 OS, so some form of minimum version control is required. This has also been discussed with @bryanhiggins.

### DIFF
--- a/api/blackberry_payment.js
+++ b/api/blackberry_payment.js
@@ -60,7 +60,7 @@ blackberry.payment = {
      *  }
      *
      *  function onFailure(err) {
-     *      alert("Error occurred: " + err.errorText + ", " + err.errorID);
+     *      alert("Error occurred: " + err.errorText + ", " + err.errorCode + ", " + err.errorInfo);
      *  }
      * &lt;/script&gt;
      */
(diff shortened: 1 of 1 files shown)


api/blackberry_payment.js
```diff
@@ -477,24 +477,43 @@ Purchase.prototype.extraParameters = null;
...
+ * on the error that occurred.
+ * @BB10X
+ *
+ */
+PaymentError.prototype.errorInfo = null;
+
+/**
  * @type String
  * @description The message for the particular error. In addition to descriptive text, error code may appear at the end of the message.
  * @BB10X
```
scdowney, line 516: Is there a specific marker for 10.2?

bryanhiggins:
@scdowney I just created a "cordova" branch to capture these changes. Could you please close this and re-issue against that branch.

PR #476, opened 2013-11-08
Spelling

jsoref:
Two of these seem to cover issues which I think we thought we had fixed -- which is rather confusing but...


bryanhiggins:
@jsoref you should be working on the webworks2 branch. master is the pre-conversion legacy docs

jsoref:
Oh. Brother.

Thanks.

